PR #5, opened 2015-03-17
Add an end2end test using a class.

mprobst:
Admittedly this is mostly to test travis.

### DIFF
--- a/test/e2e/helloworld.ts
+++ b/test/e2e/helloworld.ts
@@ -1,5 +1,17 @@
 import t = require("unittest/unittest");
 
+class MyClass {
+  field: string;
+
+  MyClass(someVal: string) { this.field = someVal; }
+
+  getField(): string { return this.field + " world"; }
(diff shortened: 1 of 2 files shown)


test/e2e/helloworld.ts
```diff
@@ -1,5 +1,17 @@
...
+class MyClass {
+  field: string;
+
+  MyClass(someVal: string) { this.field = someVal; }
+
+  getField(): string { return this.field + " world"; }
+}
+
 function main(): void {
   t.test("bigifies text", function() { t.expect("hello".toUpperCase(), t.equals("HELLO")); });
```
alexeagle: nit: I don't think the e2e test needs to be additive in terms of test cases or assertions. Just keep making it one bigger, more complex program. So we could replace line 12 rather than adding

mprobst: Done.

PR #6, opened 2015-03-17
Make this work as a node package

alexeagle:


mprobst:
Excuse my ignorance. Isn't it possible to build on "npm publish", so that we don't have to check in generated files?

alexeagle:
I dunno either , looking...

alexeagle:
Yes, there is a prepublish script which they mention using for
CoffeeScript, looks like just the thing. PTAL

PR #8, opened 2015-03-18
Switch language to dart, so that a DartVM is available.

mprobst:
Some version of NodeJS is available in every package, so we can just rely on
npm & node being available in the path, and rather depend on Dart. This should
probably be faster than downloading and installing Dart on every test run.

Also remove the other node versions, they don't add value for the moment.


alexeagle:
LGTM

PR #19, opened 2015-03-18
Support the 'import .. from ..' TS syntax.

alexeagle:
This is the first error I currently run into when running the transpiler on angular code.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,7 +1,7 @@
 require('source-map-support').install();
 
 var formatter = require('gulp-clang-format');
-var fs = require('fs');
+var fsx = require('fs-extra');
 var gulp = require('gulp');
 var gutil = require('gulp-util');
 var merge = require('merge2');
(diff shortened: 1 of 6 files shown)


gulpfile.js
```diff
@@ -80,25 +80,33 @@ gulp.task('test.e2e', ['test.compile'], function(done) {
```
mprobst: Not necessarily now, but shouldn't we move this over to something that globs all *_test.ts (or *_e2e_test.ts or so), and considers all the other files to be imported libraries/support code for those tests? I don't think putting everything into one file will scale medium term.

main.ts
```diff
@@ -62,6 +62,16 @@ class Translator {
     this.emit(')');
   }
 
+  visitExternalModuleReferenceExpr(expr: ts.Expression) {
+    // TODO: what if this isn't a string literal?
+    var moduleName = <ts.StringLiteralExpression>expr;
+    if (moduleName.text.lastIndexOf("./", 0) !== 0) {
```
mprobst: Is this funky code for saying "does not start with"? What about "string.startsWith" from ES6, or just !string.match(/^.\//) ?

main.ts
```diff
@@ -404,6 +414,37 @@ class Translator {
         this.emit('}');
         break;
 
+      case ts.SyntaxKind.ImportDeclaration:
+        var importDecl = <ts.ImportDeclaration>node;
+        this.emit('import');
+
```
mprobst: Nit: I'd rather not put too much vertical space into these blocks here, it's a bit unwieldy already.

main.ts
```diff
@@ -404,6 +414,37 @@ class Translator {
...
+        this.visitExternalModuleReferenceExpr(importDecl.moduleSpecifier);
+
+        if (importDecl.importClause) {
+          this.emit('show');
+          this.visit(importDecl.importClause);
+        }
```
mprobst: What's the semantic of "import 'foo';" in ES6? In Dart, this imports all symbols into the global scope, which I think doesn't match what happens in Dart (where it barfs all over your global scope). I think we should outlaw that and just fail for import without an import clause.

test/DartTest.ts
```diff
@@ -276,9 +276,16 @@ describe('transpile to dart', () => {
...
   describe('imports', () => {
-    it('translates simple import statements', () => {
+    it('translates import equals statements', () => {
       expectTranslate('import x = require("y");').to.equal(' import "package:y.dart" as x ;');
     });
+    it('translates import from statements', () => {
+      expectTranslate('import {x} from "y";').to.equal(' import "package:y.dart" show x ;');
+    });
+    it('allows import dart file from relative path', () => {
+        expectTranslate('import x = require("./y")').to.equal(' import "./y.dart" as x ;');
```
mprobst: indentation?

main.ts
alexeagle: Yes.
startsWith isn't available, we're not currently running node with --harmony. Probably should do, since we are using target: ts.ScriptTarget.ES6
For now, using the regex approach.

gulpfile.js
alexeagle: Yes, this needs to evolve further. I also think we should pull the unittest dep into a separate file (maybe code-gen'd) so we can also execute the tests in TypeScript.

main.ts
alexeagle: done

test/DartTest.ts
alexeagle: will re-format

main.ts
alexeagle: done. Better to start restrictive and fail fast, so we discover which of our assumptions are invalid.

mprobst:
LGTM.

PR #22, opened 2015-03-18
Support array literal

alexeagle:


test/DartTest.ts
```diff
@@ -154,6 +154,8 @@ describe('transpile to dart', () => {
 
     it('translates regexp literals',
        () => { expectTranslate('/wo\\/t?/').to.equal(' /wo\\/t?/ ;'); });
+
+    it('translates array literals', () => { expectTranslate('[1,2]').to.equal(' [ 1 , 2 ] ;'); });
```
mprobst: Could you add a test with a trailing comma? [1, 2, 3,]?

alexeagle: Yes, also added one with empty array.

PR #23, opened 2015-03-18
chore(build): put preinstall steps into preinstall script

chalin:


package.json
```diff
@@ -24,6 +24,7 @@
     "which": "^1.0.9"
   },
   "scripts": {
+    "preinstall": "npm install tsd && tsd update",
```
alexeagle: tsd isn't installed global here, so it might not be on the path... keep the ./node_modules/tsd/build/cli.js instead?

chalin: Explicit path to `node_modules/.bin` isn't needed since `npm` prepends it to `PATH`. See [nmp scripts doc](https://docs.npmjs.com/misc/scripts#path). You can also see that this PR's build is green on Travis.

alexeagle: ah, neat. Good point that travis would have been broken if this were wrong. Thanks!

PR #27, opened 2015-03-18
(bugfix): commas between multiple exposed symbols in import

alexeagle:


main.ts
```diff
@@ -436,7 +436,7 @@ class Translator {
         break;
       case ts.SyntaxKind.NamedImports:
       case ts.SyntaxKind.NamedExports:
-        this.visitEach((<ts.NamedImportsOrExports>node).elements);
```
mprobst: Should we just remove visitEach? I'll send a PR.

PR #28, opened 2015-03-18
(bugfix): allow untyped class member fields

alexeagle:


test/DartTest.ts
```diff
@@ -187,6 +191,8 @@ describe('transpile to dart', () => {
       expectTranslate('if (x) 1;').to.equal(' if ( x ) 1 ;');
       expectTranslate('if (x) 1; else 2;').to.equal(' if ( x ) 1 ; else 2 ;');
     });
+    it('translates throw', () => {expectTranslate('throw new Error("oops")')
```
mprobst: is this what clang-format produces?

alexeagle: Double checked: yes.

mprobst: Ugh. I'll take a look.

PR #29, opened 2015-03-18
Fix a bug with separating multiple variable declarations in a list

mprobst:
Also add a test case for multiple field declarations in a class.


mprobst:
FYI I considered removing the visitEach method, but it is actually useful, overall.

PR #31, opened 2015-03-18
Support enums

### LINKED ISSUE #10: Support enums
Both Dart and TypeScript have them. Basic syntax is the same.

chalin:
Fixes #10.

### DIFF
--- a/test/DartTest.ts
+++ b/test/DartTest.ts
@@ -1,4 +1,5 @@
 /// <reference path="../typings/chai/chai.d.ts"/>
+/// <reference path="../typings/mocha/mocha.d.ts"/>
 /// <reference path="../typings/source-map-support/source-map-support.d.ts"/>
 
 import sms = require('source-map-support');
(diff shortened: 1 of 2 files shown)


chalin:
FYI. Consider the following [TS example](http://www.typescriptlang.org/Handbook#basic-types-enum):
```typescript
enum Color {Red, Green, Blue};
var colorName: string = Color[2];
```
The expression `Color[2]` would need to be expressed as `Colors.values[2].toString()` in Dart. If this feature is to be used, then some code will need to be added to the TS/Dart facade (probably `angular/modules/angular2/src/facade/lang.{es6,dart}`. Note that `Color["Red"]` yielding 0 is supported in TS but not directly in Dart.

test/DartTest.ts
```diff
@@ -107,6 +108,24 @@ describe('transpile to dart', () => {
...
+    it('does not support empty enum', () => {
+      chai.expect(() => translateSource('enum Empty { }'))
+          .to.throw('empty enums are not supported');
+    });
+    it('does not support enum with initializer', () => {
+      chai.expect(() => translateSource('enum Color { Red = 1, Green, Blue = 4 }'))
+          .to.throw('enum initializers are not supported');
+    });
+    it('should support switch over enum', () => {
+      expectTranslate('switch(c) { case Color.Red: break; default: break; }')
```
mprobst: Just for my information, this is not different in any kind from a plain switch over anything else, is it?

test/DartTest.ts
```diff
@@ -107,6 +108,24 @@ describe('transpile to dart', () => {
...
+  describe('enums', () => {
+    it('should support basic enum declaration', () => {
+      expectTranslate('enum Color { Red, Green, Blue }')
+          .to.equal(' enum Color { Red , Green , Blue }');
+    });
+    it('does not support empty enum', () => {
+      chai.expect(() => translateSource('enum Empty { }'))
+          .to.throw('empty enums are not supported');
+    });
+    it('does not support enum with initializer', () => {
```
mprobst: Could you add a test for const enums?

test/DartTest.ts
chalin: Right. (This test just made me feel better that a case label could be written as, e.g., `Color.Red`. I didn't know for sure until I wrote the test.)

test/DartTest.ts
chalin: Could we address that in a separate issue / PR? Support or not of `const enum`s is IMHO debatable (in particular since there is no equivalent feature in Dart.

mprobst: Sure. I'm also good with just failing for const enums for the time being.

PR #37, opened 2015-03-19
Handle super() calls.

mprobst:
super() calls in constructors are special, they have to be turned into
an initializer call.
super.foo() style calls elsewhere do not need special treatment and are
translated as is.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -31,10 +31,14 @@ class Translator {
     }
   }
 
-  visitFunctionLike(fn: ts.FunctionLikeDeclaration) {
+  visitParameters(fn: ts.FunctionLikeDeclaration) {
     this.emit('(');
     this.visitList(fn.parameters);
     this.emit(')');
+  }
+
+  visitFunctionLike(fn: ts.FunctionLikeDeclaration) {
+    this.visitParameters(fn);
     this.visit(fn.body);
   }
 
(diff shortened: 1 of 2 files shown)


test/DartTest.ts
```diff
@@ -306,6 +306,18 @@ describe('transpile to dart', () => {
     it('translates "void"', () => {
       chai.expect(() => translateSource('void x;')).to.throw('void operator is unsupported');
     });
+    it('translates "super()" constructor calls', () => {
+      expectTranslate('class X { constructor() { super(1); } }')
+          .to.equal(' class X { X ( ) : super ( 1 ) { /* super call moved to initializer */ ; } }');
+      expectTranslate('class X { constructor() { x(); if (y) super(1, 2); z(); } }')
+          .to.equal(' class X { X ( ) : super ( 1 , 2 ) { x ( ) ;' +
+                    ' if ( y ) /* super call moved to initializer */ ;' +
```
alexeagle: That is a troubling change in behavior. Did you find an example of how the current transpiler handles this case?

mprobst: Yes, I first had this only supporting super() calls as the very first child of the ctor method, then changed it to this and didn't feel to great about it.

The Traceur transpiler handles super() calls anywhere in the body but not in further children; I've  changed this to match that behaviour. We could reconsider or print a warning if it turns out to be problematic.

main.ts
```diff
@@ -72,6 +76,44 @@ class Translator {
     this.visit(expr);
   }
 
+  /** Searches for super() calls and emits an initializer for it if found. */
+  maybeEmitInitializerForSuperCall(body: ts.Block) {
```
alexeagle: nit: naming: this is emitting the colon, and therefore the entire initializer list, which could include field initializers.

main.ts
```diff
@@ -72,6 +76,44 @@ class Translator {
...
+  /**
+   * Checks whether `callExpr` is a super() call that should be ignored because it was already
+   * handled by `maybeEmitSuperInitializer` above.
+   */
+  isHandledSuperCall(callExpr: ts.CallExpression): boolean {
+    if (callExpr.expression.kind !== ts.SyntaxKind.SuperKeyword) return false;
+    // Sanity check that there was indeed a ctor directly above this call.
+    var exprStmt = callExpr.parent;
+    var ctorBody = exprStmt.parent;
+    // TODO(martinprobst): Warn if exprStmt is not the first element of ctorBody.expressions to call
```
alexeagle: opt: yeah that would be nice

main.ts
```diff
@@ -72,6 +76,44 @@ class Translator {
...
+    };
+
+    body.statements.forEach(findSuperCalls);
+  }
+
+  /**
+   * Checks whether `callExpr` is a super() call that should be ignored because it was already
+   * handled by `maybeEmitSuperInitializer` above.
+   */
+  isHandledSuperCall(callExpr: ts.CallExpression): boolean {
```
alexeagle: does this make it easy for the two methods to get out-of-sync? like we make a change above so some expression is now a handled super call, but forget to update this method to also return true for that case?
Maybe instead this could return the expressions discovered in the ctorBody which should be produced as initializers?

main.ts
mprobst: done.

main.ts
mprobst: we'll have to rework our error/warning system at some point anyway, let's do that then.

alexeagle: actually, the easy thing to do would be to emit a comment in the dart code explaining what the transpiler did. that's the only output users should need to read.

main.ts
mprobst: There's two things that have to happen:
1) when you hit the ctor, you find the super call and move it to the initializer
2) when you hit a super() call, you have to find if it should be removed or error out

(2) has to happen independently of (1), as you could find a super call anywhere in the tree.

So I think this is the best we can do here.

main.ts
mprobst: That's exactly what this does, isn't it?

PR #38, opened 2015-03-19
Temporarily use a TypeScript that includes parent pointer support.

mprobst:


mprobst:
This is superseded by #43 (which includes this commit).

PR #39, opened 2015-03-19
Line and character offset are 0 based, so add 1.

mprobst:


mprobst:
The build failures are a fluke, I believe.

PR #40, opened 2015-03-19
Only transpile if the script is the entry point.

mprobst:
Also be more selective about which files to pick
up in typings, as our TS compiler now accepts all
file types, not just .ts files.


mprobst:
@alexeagle FYI, merging this now.

alexeagle:
LGTM

PR #41, opened 2015-03-19
Support interfaces and abstract methods in them.

mprobst:


test/DartTest.ts
```diff
@@ -113,6 +113,9 @@ describe('transpile to dart', () => {
       expectTranslate('class X extends Y implements Z {}')
           .to.equal(' class X extends Y implements Z { }');
     });
+    it('supports abstract methods', () => {
+      expectTranslate('interface X { x(); }').to.equal(' abstract class X { abstract x ( ) ; }');
```
alexeagle: opt: we should add to the e2e test to exercise the subtyping/subclassing relations and how they work at runtime

alexeagle:
curious: did you find anything in the angular codebase which required this?

main.ts
```diff
@@ -446,6 +450,14 @@ class Translator {
         this.visitFunctionLike(funcExpr);
         break;
 
+      case ts.SyntaxKind.MethodSignature:
+        var methodSignatureDecl = <ts.FunctionLikeDeclaration>node;
+        this.emit('abstract');
+        if (methodSignatureDecl.modifiers) this.visitEach(methodSignatureDecl.modifiers);
```
alexeagle: maybe visitEach should short-circuit if the argument is falsy

mprobst: Yeah. I'll add a maybeVisitEach for the cases where stuff is nullable. I don't like methods that are too acceptant of their args, so having a strict and a maybe variant might be nice.

PR #43, opened 2015-03-19
Parent nodes

mprobst:


alexeagle:
Any reason to switch repos? Did the needed change not show up at head in ms repo?

mprobst:
Mohamed has submitted the code, but they haven't built a release (which doesn't happen automatically with npm in their setup). So this is just temporary until there's a new TS release.

PR #47, opened 2015-03-19
Refactor super call handling to be more local.

mprobst:
Handle the emit() operation in the super call detector itself.


mprobst:
@alexeagle FYI.

PR #50, opened 2015-03-19
Fail on const enums

### LINKED ISSUE #35: Support const enums
Possibly by just failing to translate them if there's no reasonable equivalent in Dart.

chalin:
Fixes #35.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -474,6 +474,11 @@ class Translator {
 
       case ts.SyntaxKind.EnumDeclaration:
         var decl = <ts.EnumDeclaration>node;
+        // The only legal modifier for an enum decl is const.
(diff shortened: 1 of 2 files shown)


main.ts
```diff
@@ -414,6 +414,10 @@ class Translator {
 
       case ts.SyntaxKind.EnumDeclaration:
         var decl = <ts.EnumDeclaration>node;
+        // The only legal modifier for an enum decl is const.
```
mprobst: Could you be explicit about this containing const? It might change in the future, and it's also more readable. I think "export enum Foo" might also be a thing.

chalin: Done.

PR #60, opened 2015-03-20
Support & check visibility modifiers.

### LINKED ISSUE #44: Require that private members have prefix _, public ones do not
It should be a transpiler error if a class member is declared to be `private` and its name does not begin with `_`, and the inverse for `public` members:

```
class Foo {
  private _bar;  // valid
  baz;  // valid
  private qux;  // invalid
  _quux;  // invalid
}
```

mprobst:
Fixes #44.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -117,6 +117,24 @@ class Translator {
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


yjbanov:
lgtm

Issue #72, opened 2015-03-20
Support Annotations


mprobst:
It's unclear why they are currently not properly parsed.

PR #77, opened 2015-03-21
Initial support for decorators.

mprobst:
This is based on a hacked up version of TypeScript including @mhegazy's minimal decorators pull request. I.e. this is to allow further progress on ts2dart transpilation, not a final solution.

So this relates to #72, but doesn't fix it.


alexeagle:
LGTM

PR #89, opened 2015-03-23
Multiple variable initializers in for loop

### LINKED ISSUE #83: Multiple variable initializers in for loop (`for (var x, y = 0; x; x++) { .. }`)

chalin:
Fixes #83.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -190,6 +190,32 @@ class Translator {
     }
   }
 
+  visitVariableDeclarationType(varDecl: ts.VariableDeclaration) {
+    /* Note: VariableDeclarationList can only occur as part of a for loop. This helper method
+     * is meant for processing for-loop variable declaration types only.
+     *
+     * In Dart, all variables in a variable declaration list must have the same type. Since
+     * we are doing syntax directed translation, we cannot reliably determine if distinct
+     * variables are declared with the same type or not. Hence we support the following cases:
+     *
+     * - A variable declaration list with a single variable can be explicitly typed.
+     * - When more than one variable is in the list, all must be implicitly typed.
+     */
+    var firstDecl = varDecl.parent.declarations[0];
+    var msg = 'Variables in a declaration list of more than one variable cannot by typed';
+    if (firstDecl === varDecl) {
+      if (!varDecl.type) {
+        this.emit('var');
+      } else if (varDecl.parent.declarations.length > 1) {
+        this.reportError(varDecl, msg);
+      } else {
+        this.visit(varDecl.type);
+      }
+    } else if (varDecl.type) {
+      this.reportError(varDecl, msg);
+    }
+  }
+
   reportError(n: ts.Node, message: string) {
     var file = n.getSourceFile() || this.currentFile;
     var start = n.getStart(file);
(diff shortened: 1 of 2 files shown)


test/DartTest.ts
```diff
@@ -41,8 +41,19 @@ describe('transpile to dart', () => {
       expectTranslate('var a;').to.equal(' var a ;');
       expectTranslate('var a:any;').to.equal(' dynamic a ;');
     });
-    it('should transpile variable declaration lists',
-       () => { expectTranslate('var a: number, b: string;').to.equal(' num a ; String b ;'); });
+    it('should transpile variable declaration lists', () => {
+      expectTranslate('var a, b;').to.equal(' var a , b ;');
+      expectTranslate('var a: number, b;').to.equal(' num a , b ;');
```
mprobst: doesn't this mean that 'b' is also a num in Dart?

main.ts
```diff
@@ -217,13 +217,17 @@ class Translator {
...
+        // Note that a VariableDeclarationList can only occur as part of a for loop.
         var varDeclList = <ts.VariableDeclarationList>node;
-        this.visitList(varDeclList.declarations, ';');
+        this.visitList(varDeclList.declarations, ',');
         break;
 
       case ts.SyntaxKind.VariableDeclaration:
         var varDecl = <ts.VariableDeclaration>node;
-        if (varDecl.type) {
+        if (varDecl.parent.declarations[0] !== varDecl) {
```
mprobst: maybe it'd be more readable to put this check into the `if (varDecl.type)` branch below?

test/DartTest.ts
chalin: Yes, as in other languages (C, Java, C#). I thought that that it is what it meant in TS, but apparently not. Very good catch!

chalin:
Here is a commit which implements the assumptions presented in https://github.com/angular/ts2dart/issues/83#issuecomment-85282233.

main.ts
```diff
@@ -217,16 +217,32 @@ class Translator {
         break;
 
       case ts.SyntaxKind.VariableDeclarationList:
+        // Note that a VariableDeclarationList can only occur as part of a for loop.
         var varDeclList = <ts.VariableDeclarationList>node;
-        this.visitList(varDeclList.declarations, ';');
+        this.visitList(varDeclList.declarations, ',');
```
mprobst: the comma is the default separator in list, so you can just not specify the second param.

main.ts
```diff
@@ -217,16 +217,32 @@ class Translator {
...
+         * All variables in a variable declaration list must have the same type. Since
+         * we are doing syntax directed translation, we cannot reliably determine if distinct
+         * variables are declared with the same type or not. Hence we support the following cases:
+         * - A variable declaration list with a single variable can be explicitly typed.
+         * - When more than one variable is in the list, all must be implicitly typed.
+         */
+        var firstDecl = varDecl.parent.declarations[0];
+        var msg = 'Variables in a declaration list of more than one variable cannot by typed';
+        if (firstDecl === varDecl) {
+          if (!varDecl.type) {
```
mprobst: Could you extract this to a method "visitVarDeclType"?

Then you could also use early exits from the function to make the boolean conditions a bit easier to read.

test/DartTest.ts
```diff
@@ -41,8 +41,21 @@ describe('transpile to dart', () => {
...
+    it('should transpile variable declaration lists with initializers', () => {
+      expectTranslate('var a = 0;').to.equal(' var a = 0 ;');
+      expectTranslate('var a, b = 0;').to.equal(' var a , b = 0 ;');
+      expectTranslate('var a = 1, b = 0;').to.equal(' var a = 1 , b = 0 ;');
+    });
+    it('does not support declaration lists containing more than one type (implicit or explicit)', () => {
+      var msg = 'Variables in a declaration list of more than one variable cannot by typed';
+      expectErroneousCode('var a: A, b;').to.throw(msg);
+      expectErroneousCode('var a, b: B;').to.throw(msg);
+      expectErroneousCode('var a: number, b: string;').to.throw(msg);
```
mprobst: Nice tests. I'd add one more case:
```var untyped, a: number, b: string;```

main.ts
chalin: Good suggestion.

chalin:
Changes done. Note that the new visit method has signature `visitVariableDeclarationType(varDecl: ts.VariableDeclaration)` rather than `visitVariableDeclarationType(t: ts.TypeNode)` because a parent cannot be extracted from `t` (to get the `varDecl`) when it is absent.

mprobst:
I've merged this in 27eaa56.

PR #90, opened 2015-03-24
support destructuring params with defaults.

### LINKED ISSUE #81: Default values for named parameters ({x} = {}) incorrectly handled (angular2/src/core/annotations/annotations.dart)

alexeagle:
Fixes #81

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -122,6 +122,21 @@ class Translator {
     return true;
   }
 
+  visitNamedParameter(paramDecl: ts.ParameterDeclaration) {
+    this.visitEachIfPresent(paramDecl.decorators);
+    if (paramDecl.type) {
+      // TODO(martinprobst): These are currently silently ignored.
+      // this.reportError(paramDecl.type, 'types on named parameters are unsupported');
+    }
+    this.visit(paramDecl.name);
+    if (paramDecl.initializer) {
+      if (paramDecl.initializer.kind !== ts.SyntaxKind.ObjectLiteralExpression
+        || (<ts.ObjectLiteralExpression>paramDecl.initializer).properties.length > 0) {
+        this.reportError(paramDecl, 'initializers for named parameters must be empty object literals');
+      }
+    }
+  }
+
   visitExternalModuleReferenceExpr(expr: ts.Expression) {
     // TODO: what if this isn't a string literal?
     var moduleName = <ts.StringLiteralExpression>expr;
@@ -698,16 +713,13 @@ class Translator {
       case ts.SyntaxKind.Parameter:
         var paramDecl = <ts.ParameterDeclaration>node;
         if (paramDecl.dotDotDotToken) this.reportError(node, 'rest parameters are unsupported');
+        if (paramDecl.name.kind === ts.SyntaxKind.ObjectBindingPattern) {
+          this.visitNamedParameter(paramDecl);
+          break;
+        }
         if (paramDecl.initializer) this.emit('[');
         this.visitEachIfPresent(paramDecl.decorators);
-        if (paramDecl.type) {
-          if (paramDecl.name.kind !== ts.SyntaxKind.ObjectBindingPattern) {
-            this.visit(paramDecl.type);
-          } else {
-            // TODO(martinprobst): These are currently silently ignored.
-            // this.reportError(paramDecl.type, 'types on named parameters are unsupported');
-          }
-        }
+        if (paramDecl.type) this.visit(paramDecl.type);
         this.visit(paramDecl.name);
         if (paramDecl.initializer) {
           this.emit('=');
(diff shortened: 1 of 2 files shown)


main.ts
```diff
@@ -699,21 +699,23 @@ class Translator {
...
             // TODO(martinprobst): These are currently silently ignored.
             // this.reportError(paramDecl.type, 'types on named parameters are unsupported');
           }
-        }
-        this.visit(paramDecl.name);
-        if (paramDecl.initializer) {
-          this.emit('=');
-          this.visit(paramDecl.initializer);
-          this.emit(']');
+          this.visit(paramDecl.name);
```
mprobst: I think this code should check that the initialiser is just an empty object assignment, and if not, report an error.

main.ts
```diff
@@ -699,21 +699,23 @@ class Translator {
       case ts.SyntaxKind.Parameter:
         var paramDecl = <ts.ParameterDeclaration>node;
         if (paramDecl.dotDotDotToken) this.reportError(node, 'rest parameters are unsupported');
-        if (paramDecl.initializer) this.emit('[');
-        this.visitEachIfPresent(paramDecl.decorators);
-        if (paramDecl.type) {
-          if (paramDecl.name.kind !== ts.SyntaxKind.ObjectBindingPattern) {
-            this.visit(paramDecl.type);
-          } else {
+        if (paramDecl.name.kind === ts.SyntaxKind.ObjectBindingPattern) {
```
mprobst: This is sort of beyond the threshold of a single case block, isn't it? Maybe extract to a separate function?

test/DartTest.ts
```diff
@@ -206,12 +206,16 @@ describe('transpile to dart', () => {
       expectErroneousCode('function x({a}: number) { return a + b; }')
           .to.throw('types on named parameters are unsupported');
     });
+    it('translates destructuring parameters', () => {
+      expectTranslate('function x({p = null, d = false} = {}) {}')
```
mprobst: add a test to fail if the initializer is more than just an empty object?

main.ts
alexeagle: yes, good idea

main.ts
alexeagle: extracted a function for the unusual case.
This file is getting unmanageable, the distance from this point to the helper function is large. don't have a suggestion what to do about it...

test/DartTest.ts
alexeagle: done

main.ts
```diff
@@ -122,6 +122,23 @@ class Translator {
...
+      // TODO(martinprobst): These are currently silently ignored.
+      // this.reportError(paramDecl.type, 'types on named parameters are unsupported');
+    }
+    this.visit(paramDecl.name);
+    if (paramDecl.initializer) {
+      if (paramDecl.initializer.kind !== ts.SyntaxKind.ObjectLiteralExpression) {
+        this.reportError(paramDecl, 'named parameter initializer must be empty object literal');
+      }
+      if ((<ts.ObjectLiteralExpression>paramDecl.initializer).properties.length > 0) {
+        this.reportError(paramDecl, 'non-empty named parameter initializers are unsupported');
```
mprobst: I'd say the same as above, `initializers for named parameters must be empty object literals`, fewer negations and tells you what to do.

test/DartTest.ts
```diff
@@ -165,27 +164,21 @@ describe('transpile to dart', () => {
...
-    it('translates on functions', () => {
-      expectTranslate('@A function f() {}').to.equal(' @ A f ( ) { }');
-    });
-    it('translates on properties', () => {
-      expectTranslate('class X { @A p; }').to.equal(' class X { @ A var p ; }');
-    });
-    it('translates on parameters', () => {
-      expectTranslate('function f (@A p) {}').to.equal(' f ( @ A p ) { }');
-    });
+    it('translates plain decorators',
```
mprobst: could you split out these changes? It's kind of hard to figure out what actually changed here.

main.ts
```diff
@@ -122,6 +122,23 @@ class Translator {
...
   }
 
+  visitNamedParameter(paramDecl: ts.ParameterDeclaration) {
+    this.visitEachIfPresent(paramDecl.decorators);
+    if (paramDecl.type) {
+      // TODO(martinprobst): These are currently silently ignored.
+      // this.reportError(paramDecl.type, 'types on named parameters are unsupported');
+    }
+    this.visit(paramDecl.name);
+    if (paramDecl.initializer) {
```
mprobst: actually that's going sort of towards a linter, but should this also fail if there is no initializer for a named parameter list? That's AFAICT always wrong (fails if you don't pass any named parameters).

main.ts
alexeagle: done, and added another test assertion

test/DartTest.ts
alexeagle: hey, no fair! if I run clang-format and get spurious deltas, that's the fault of whoever commited previously.
I'll send a format-only change and rebase this on top.

main.ts
alexeagle: I think you're saying there's another validation we could add? I didn't quite follow. Maybe you're saying there's a bug in this code?

mprobst:
LGTM.

PR #92, opened 2015-03-24
Run clang-format on the code.

alexeagle:
We need a better mechanism to prevent regressions. Should the check-format task be configured to fail now? It would be nicer to fail later in the dev workflow, like a separate warning on the PR

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -30,7 +30,7 @@ class Translator {
 
   visitEach(nodes: ts.Node[]) { nodes.forEach((n) => this.visit(n)); }
 
-  visitEachIfPresent(nodes?: ts.Node[]) {
+  visitEachIfPresent(nodes ?: ts.Node[]) {
     if (nodes) this.visitEach(nodes);
   }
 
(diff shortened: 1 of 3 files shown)


main.ts
```diff
@@ -30,7 +30,7 @@ class Translator {
 
   visitEach(nodes: ts.Node[]) { nodes.forEach((n) => this.visit(n)); }
 
-  visitEachIfPresent(nodes?: ts.Node[]) {
+  visitEachIfPresent(nodes ?: ts.Node[]) {
```
alexeagle: I imagine you'll want to track this down in clang-format?

main.ts
```diff
@@ -399,15 +399,15 @@ class Translator {
         break;
       case ts.SyntaxKind.TemplateTail:
         this.result += `${
```
alexeagle: ouch! this is a clang-format bug - it's adding the line break here.
I guess it doesn't know about backtick string literals?

Issue #82, opened 2015-03-21
Handle parameters with default values at start of parameter list (angular2/src/core/compiler/selector.dart)


mprobst:

alexeagle:
Looks like this is related to #58 or will have merge conflicts with it at least

mprobst:
Fixed by #93.

PR #93, opened 2015-03-24
Support multiple optional positional parameters.

### LINKED ISSUE #58: Cannot have multiple groups of positional parameters in a single parameter list
input:
 constructor(bindings:List, parent=null, defaultBindings=false) {

line 1, column 966 of dist/dart.ts2dart/injector.dart: Cannot have multiple groups of positional parameters in a single parameter list
...
Injector ( List bindings , [ parent = null ] , [ defaultBindings = false ] )

mprobst:
Fixes #58 and #82.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -43,7 +43,28 @@ class Translator {
 
   visitParameters(fn: ts.FunctionLikeDeclaration) {
     this.emit('(');
-    this.visitList(fn.parameters);
+    let i;
+    for (i = 0; i < fn.parameters.length; i++) {
+      // ObjectBindingPatterns are handled within the parameter visit.
+      if (fn.parameters[i].initializer &&
+          fn.parameters[i].name.kind !== ts.SyntaxKind.ObjectBindingPattern) {
+        break;
+      }
+    }
+
+    if (i !== 0) {
+      var requiredParams = fn.parameters.slice(0, i);
+      this.visitList(requiredParams);
+    }
+
+    if (i !== fn.parameters.length) {
+      if (i !== 0) this.emit(',');
+      var positionalOptional = fn.parameters.slice(i, fn.parameters.length);
+      this.emit('[');
+      this.visitList(positionalOptional);
+      this.emit(']');
+    }
+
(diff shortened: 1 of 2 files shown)


main.ts
```diff
@@ -43,7 +43,35 @@ class Translator {
 
   visitParameters(fn: ts.FunctionLikeDeclaration) {
     this.emit('(');
-    this.visitList(fn.parameters);
+    var i;
+    for (i = 0; i < fn.parameters.length; i++) {
```
alexeagle: can you use `let`?

alexeagle: I think you can, because we claim to be targetting ES6, however, we are not calling node with --harmony which you'd need to fix first.

test/DartTest.ts
```diff
@@ -185,8 +185,13 @@ describe('transpile to dart', () => {
     it('supports declarations',
        () => { expectTranslate('function x() {}').to.equal(' x ( ) { }'); });
     it('supports param default values', () => {
-      expectTranslate('function x(a = 42) { return 42; }')
-          .to.equal(' x ( [ a = 42 ] ) { return 42 ; }');
+      expectTranslate('function x(a = 42, b = 1) { return 42; }')
+          .to.equal(' x ( [ a = 42 , b = 1 ] ) { return 42 ; }');
+      expectTranslate('function x(p1, a = 42, b = 1) { return 42; }')
+          .to.equal(' x ( p1 , [ a = 42 , b = 1 ] ) { return 42 ; }');
+      expectErroneousCode('function x(a = 42, b = 1, p2) { return 42; }')
```
alexeagle: this scenario comes up in #82 and I think it needs to be allowed.

main.ts
mprobst: I can use let, it simply gets turned into var here (which is correct).

test/DartTest.ts
mprobst: `f(a, [b = 1, c = 2], d) { }` is incorrect Dart code.

PR #95, opened 2015-03-25
Correctly escape escape sequences in template literals.

### LINKED ISSUE #79: Escape \u sequences properly (angular2/src/change_detection/parser/lexer.dart)

mprobst:
Fixes #79.

### DIFF
--- a/test/DartTest.ts
+++ b/test/DartTest.ts
@@ -251,6 +251,8 @@ describe('transpile to dart', () => {
     it('translates fat arrow operator', () => {
       expectTranslate('var a = () => {}').to.equal(' var a = ( ) { } ;');
       expectTranslate('var a = (p) => isBlank(p)').to.equal(' var a = ( p ) => isBlank ( p ) ;');
+      expectTranslate('var a = (p = null) => isBlank(p)')
+          .to.equal(' var a = ( [ p = null ] ) => isBlank ( p ) ;');
     });
   });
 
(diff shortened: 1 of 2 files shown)


test/DartTest.ts
```diff
@@ -275,6 +277,10 @@ describe('transpile to dart', () => {
       expectTranslate("`\\$a`").to.equal(" '''\\$a''' ;");
     });
 
+    it('escapes escape sequences', () => {
+      expectTranslate("`\\\\u1234`").to.equal(" '''\\\\u1234''' ;");
```
alexeagle: but we still have a problem in the angular code, right? https://github.com/angular/angular/blob/master/modules/angular2/src/change_detection/parser/lexer.js#L375 doesn't have four backslashes in the file, it has two. Unless the problem is that we have to escape the input here in a way we don't when reading from a file?

mprobst: No, we're all good. This has four backslashes, because it's TypeScript source that is escaped within a TypeScript string, doubling the number of backslashes. The actual string after parsing has two backslashes, like the Angular source. And that then generates a two-backslash result, as expected and tested for.

Yeah, this is all a bit confusing.

alexeagle: Yes, I believe you now. The extra double quotes around the string are the clue.
An e2e test would be easier to read in this respect, though I think this edge case doesn't need such a heavyweight verification.

PR #96, opened 2015-03-25
Handle paths in exports just like in imports.

### LINKED ISSUE #17: Avoid leading "./" when importing/exporting from a subdirectory
```dart
library angular2.directives_dart;

export './src/directives/foreach.dart';
export './src/directives/if.dart';
export './src/directives/non_bindable.dart';
export './src/directives/switch.dart';
```

should be
```dart
library angular2.directives_dart;

export 'src/directives/foreach.dart';
export 'src/directives/if.dart';
export 'src/directives/non_bindable.dart';
export 'src/directives/switch.dart';
```

mprobst:
Also strip './' to be more idiomatic Dart (Fixes #17)

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -162,10 +162,15 @@ class Translator {
   visitExternalModuleReferenceExpr(expr: ts.Expression) {
     // TODO: what if this isn't a string literal?
     var moduleName = <ts.StringLiteralExpression>expr;
-    if (!moduleName.text.match(/^\./)) {
-      moduleName.text = 'package:' + moduleName.text;
+    var text = moduleName.text;
+    if (text.match(/^\.\//)) {
+      // Strip './' to be more Dart-idiomatic.
+      text = text.substring(2);
+    } else {
+      // Unprefixed imports are package imports.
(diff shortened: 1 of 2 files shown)


main.ts
```diff
@@ -162,10 +162,13 @@ class Translator {
   visitExternalModuleReferenceExpr(expr: ts.Expression) {
     // TODO: what if this isn't a string literal?
     var moduleName = <ts.StringLiteralExpression>expr;
-    if (!moduleName.text.match(/^\./)) {
-      moduleName.text = 'package:' + moduleName.text;
+    var text = moduleName.text;
+    if (!text.match(/^\./)) {
```
alexeagle: this should probably be the same regex.
opt: could evaluate the match once and then use a capture group instead of replace below - but performance is not important here

alexeagle:
LGTM

main.ts
mprobst: Changed to just use substring(2), and inverted the options to have less negation.

kevmoo:
Awesome! :metal:

PR #99, opened 2015-03-26
Special case the @CONST, @ABSTRACT, and @IMPLEMENTS annotations.

### LINKED ISSUE #98: Handle @CONST, @IMPLEMENTS, @ABSTRACT

mprobst:
Fixes #98.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -11,6 +11,11 @@ class Translator {
   lastCommentIdx: number = -1;
   currentFile: ts.SourceFile;
   errors: string[] = [];
+  failFast: boolean;  // For tests, fail on the first problem.
+
+  constructor(failFast: boolean = false) {
+    this.failFast = failFast;
+  }
 
   translate(sourceFile: ts.SourceFile) {
     this.currentFile = sourceFile.getSourceFile();
@@ -88,7 +93,7 @@ class Translator {
   }
 
   visitClassLike(keyword: string, decl: ts.ClassDeclaration | ts.InterfaceDeclaration) {
-    this.visitEachIfPresent(decl.decorators);
+    this.visitDecorators(decl.decorators);
     this.emit(keyword);
     this.visit(decl.name);
     if (decl.typeParameters) {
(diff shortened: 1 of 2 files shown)


mprobst:
Sorry about the size of it. And overall it's pretty ugly hacks - but once we're out of AtScript, we can drop some of it again.

main.ts
```diff
@@ -97,11 +102,42 @@ class Translator {
...
+        this.emit('implements');
+      }
+      this.emit(implIfs.join(' , '));
+    }
     this.emit('{');
     this.visitEachIfPresent(decl.members);
     this.emit('}');
   }
 
+  /** Returns parameters to @IMPLEMENTS, returns their identifiers. */
```
alexeagle: this doesn't read right

main.ts
```diff
@@ -97,11 +102,42 @@ class Translator {
...
+  getImplementsDecorators(decorators: ts.NodeArray<ts.Decorator>): string[] {
+    var interfaces = [];
+    if (!decorators) return interfaces;
```
alexeagle: interesting that you don't have to declare the type of interfaces

main.ts
```diff
@@ -111,6 +147,58 @@ class Translator {
...
+        return;
+      }
+      if (name === 'IMPLEMENTS') {
+        // Ignore @IMPLEMENTS - it's handled above in visitClassLike.
+        return;
+      }
+      this.emit('@');
+      this.visit(d.expression);
+    });
+    if (isAbstract) this.emit('abstract');
```
alexeagle: could emit inside the loop, unless we are really concerned that CONST or ABSTRACT appears more than once

main.ts
```diff
@@ -174,21 +262,92 @@ class Translator {
     this.visit(expr);
   }
 
-  /** Searches for super() calls and emits an initializer for it if found. */
-  maybeEmitInitializerListForSuperCall(body: ts.Block) {
-    if (!body || body.statements.length === 0) return;
+  /**
+   * Handles constructor initializer lists and bodies.
+   *
+   * <p>Dart's super() ctor calls have to be moved to the constructors initializer list, and `const`
```
alexeagle: note
https://github.com/angular/angular/issues/70

(I don't think you're fixing here?)

test/DartTest.ts
```diff
@@ -537,5 +565,5 @@ export function translateSource(contents: string): string {
     var first = program.getSyntacticDiagnostics()[0];
     throw new Error(`${first.start}: ${first.messageText} in ${contents}`);
   }
-  return main.translateProgram(program);
+  return main.translateProgram(program, {failFast: failFast});
```
alexeagle: is this the place where you can do the shorthand {failFast} ?

alexeagle:
Nice, if ugly. LGTM

main.ts
mprobst: done.

main.ts
mprobst: That's odd indeed. I checked the spec:
https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3.10.3

This sounds like it's a bug, at least I cannot find language in the spec to the point of converting `Array<any>` to `Array<string>`.

main.ts
mprobst: Done. I don't think TS allows duplicated annotations.

main.ts
mprobst: Nope, this is just handling the initializer stuff, ctor inheritance is a different can of worms.

test/DartTest.ts
mprobst: Done. I'm still not sure I like that syntax.

mprobst:
Yeah, these AST transformations always escalate quickly. But stuff got done, which is always nice.

PR #101, opened 2015-03-26
Compile field as final if there's a const constructor.

mprobst:


test/DartTest.ts
```diff
@@ -201,8 +201,11 @@ describe('transpile to dart', () => {
     it('special cases @CONST', () => {
       expectTranslate('@CONST class X {}').to.equal(' const class X { }');
       expectTranslate('@CONST() class X {}').to.equal(' const class X { }');
-      expectTranslate('class X { @CONST constructor() { super(3); this.x = 1; this.y = 2; } }')
-          .to.equal(' class X { const X ( ) : x = 1 , y = 2 , super ( 3 ) ; }');
+      expectTranslate('class X { ' +
```
alexeagle: opt: use backtick multi-line string literal? or is that too hard to indent?

test/DartTest.ts
```diff
@@ -201,8 +201,11 @@ describe('transpile to dart', () => {
...
+                      '  x: number; ' +
+                      '  @CONST constructor() { super(3); this.x = 1; this.y = 2; } ' +
+                      '}')
```
alexeagle: nit: this would be better as a separate assertion with just the member and no statements in the ctor body

alexeagle:
LGTM

test/DartTest.ts
mprobst: Done.

test/DartTest.ts
mprobst: Well, but the result wouldn't be valid code then (doesn't initialize the final fields). I think it's fine like this.

PR #105, opened 2015-03-27
Refactor the fs layout before refactoring main.ts.

mprobst:
Move sources into a "lib" subdir (just main.ts for now).
Rename release -> build.
Produce source in build/src so that references from tests work.
Don't use a tmpdir for e2e tests.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,6 +1,7 @@
 require('source-map-support').install();
 
 var formatter = require('gulp-clang-format');
+var fs = require('fs');
 var fsx = require('fs-extra');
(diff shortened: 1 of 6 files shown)


gulpfile.js
```diff
@@ -1,6 +1,7 @@
 require('source-map-support').install();
 
 var formatter = require('gulp-clang-format');
+var fs = require('fs');
```
alexeagle: fs-extra extends fs, so you could just use fsx below
but, since that is surprising, I'm also okay with being explicit here

alexeagle:
nice, LGTM

PR #107, opened 2015-03-30
Move the translation functions onto Transpiler.

mprobst:
As more configuration and state needs to be passed, passing the state
down from three helper functions gets unmanageable.
This also fixes passing in the library name as a relative file name, and
moves the path -> library name logic into this plugin (I will create a
matching gulp-ts2dart release).

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -5,19 +5,55 @@ require('source-map-support').install();
 import ts = require("typescript");
 import fs = require("fs");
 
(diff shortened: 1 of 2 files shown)


alexeagle:
LGTM - but check travis failure

mprobst:
Oh nice, classic good old `this` pointer pointing to undefined issue. I had almost forgotten that this is JS ;-)

PR #109, opened 2015-03-30
const for collection literals in annotations.

### LINKED ISSUE #108: Arguments passed to metadata annotations/decorators must be const
E.g. note the const:

```
@Template(directives: const [ADirective])
class Foo {}
```

mprobst:
Annotation arguments must be const in Dart. Special case object and
array literals if they are somewhere within a decorator tree.

Fixes #108.

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -214,6 +214,13 @@ export class Transpiler {
     });
   }
 
+  hasAncestor(n: ts.Node, kind: ts.SyntaxKind): boolean {
+    for (var parent = n; parent; parent = parent.parent) {
+      if (parent.kind === kind) return true;
+    }
+    return false;
+  }
+
   hasAnnotation(decorators: ts.NodeArray<ts.Decorator>, name: string): boolean {
     return decorators && decorators.some((d) => {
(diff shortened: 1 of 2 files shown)


alexeagle:
This might be a nice case for e2e test

lib/main.ts
```diff
@@ -176,6 +176,13 @@ class Translator {
     });
   }
 
+  hasAncestor(n: ts.Node, kind: ts.SyntaxKind): boolean {
```
alexeagle: opt: I think there is a helper exposed by ts for this

alexeagle:
LGTM

lib/main.ts
mprobst: I can't find anything like that, AFAICT there's only `ts.forEachChild`.

mprobst:
As for the e2e test, I think that might be too hard right now. We don't have an agreed upon reflection API, so it's not possible to write code that does something with annotations (or decorators), so the integration test cannot really have behaviour that's testable. The TypeScript APIs here are also still somewhat in flux. I'll skip this for now.

PR #111, opened 2015-03-31
Support object literal shorthand expressions.

mprobst:


test/DartTest.ts
```diff
@@ -350,6 +350,7 @@ describe('transpile to dart', () => {
       expectTranslate('var x = {a: 1, b: 2}').to.equal(' var x = { "a" : 1 , "b" : 2 } ;');
       expectTranslate('var x = {a: 1, }').to.equal(' var x = { "a" : 1 } ;');
       expectTranslate('var x = {}').to.equal(' var x = { } ;');
+      expectTranslate('var x = {y}').to.equal(' var x = { "y" : y } ;');
```
yjbanov: I don't think we should support object literals. The semantics are way too different. If you want to use a map in JS, use Map. Otherwise define a class. We already have a facade to work with maps. The transpiler should simply throw in this case.

alexeagle:
LGTM aside from Yegor's point

test/DartTest.ts
mprobst: The current Angular code base cannot be translated without support for object literals, they are used in many places. They do convert to Dart maps (as they did in Traceur). This commit just adds support for the shorthand notation `{foo}` where `foo` is both the key and the value.

If we wanted to ban all usage of objects-as-maps from the Angular code base, that might be possible, but it sure would be a ton of work. Also note that JS doesn't have a literal to create maps, AFAIK, so this would be a substantial inconvenience.

PR #112, opened 2015-03-31
Fix the Angular build.

mprobst:
Ignore non .js/.ts files to avoid transpiling Angular's .es6 shims.
Fail in a more understandable way for "export {x as y}".
Bump version for a new release.


alexeagle:
When you are back, I'd like to talk more about the broken build and understand why es6 files were showing up. It uses the same config to find sources as the dart compile:
https://github.com/angular/angular/blob/master/gulpfile.js#L123
Do we know what commit caused it to go red?

Issue #113, opened 2015-03-31
Warn when TypeScript identifiers are Dart reserved words


yjbanov:
The list is available in the [spec](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-408.pdf).

PR #115, opened 2015-04-02
Refine identifying keyword kind (in prep for fix)

chalin:
* This is preparatory work for #113.
* Note that there are **three kinds of keywords**: reserved, built-in, and
limited. Only reserved words must be avoided when porting from JS to
Dart (see [footnote 1 in Dart Up and Running Chap. 2](https://www.dartlang.org/docs/dart-up-and-running/ch02.html#keywords)).
* `getLibraryName()` has been tweaked to only rename libraries having names that match a *reserved word*.
* Added test case for built-in and limited keywords.

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -497,19 +497,24 @@ export class Transpiler {
     this.emit(translated);
   }
 
-  static DART_KEYWORDS =
-      ('abstract as assert async await break case catch class const continue ' +
-       'default deferred do dynamic else enum export extends external factory false final ' +
-       'finally for get if implements import in is library new null operator part rethrow return' +
-       ' set static super switch sync this throw true try typedef var void while with yield')
-          .split(/ /);
+  // For the Dart keyword list see
+  // https://www.dartlang.org/docs/dart-up-and-running/ch02.html#keywords
+  static DART_RESERVED_WORDS =
+      ('assert break case catch class const continue default do else enum extends false final ' +
+      'finally for if in is new null rethrow return super switch this throw true try var void ' +
+      'while with').split(/ /);
+
+  // These are the built-in and limited keywords.
+  static DART_OTHER_KEYWORDS =
+      ('abstract as async await deferred dynamic export external factory get implements import ' +
+      'library operator part set static sync typedef yield').split(/ /);
 
   getLibraryName(nameForTest: string = null) {
     var parts = (nameForTest || this.relativeFileName).split('/');
     return parts.filter((p) => p.length > 0)
         .map((p) => p.replace(/[^\w.]/g, '_'))
         .map((p) => p.replace(/\.[jt]s$/g, ''))
-        .map((p) => Transpiler.DART_KEYWORDS.indexOf(p) != -1 ? '_' + p : p)
+        .map((p) => Transpiler.DART_RESERVED_WORDS.indexOf(p) != -1 ? '_' + p : p)
         .join('.');
   }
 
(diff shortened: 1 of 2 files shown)


lib/main.ts
```diff
@@ -497,19 +499,30 @@ export class Transpiler {
...
-      ('abstract as assert async await break case catch class const continue ' +
-       'default deferred do dynamic else enum export extends external factory false final ' +
-       'finally for get if implements import in is library new null operator part rethrow return' +
-       ' set static super switch sync this throw true try typedef var void while with yield')
-          .split(/ /);
+  // Dart keyword list copied from
+  // https://www.dartlang.org/docs/dart-up-and-running/ch02.html#keywords
+  // Note the different kinds of keywords defined by the enum; key values
+  // match the footnote numbers given in the docs, or 0 if none given.
+  static DART_KEYWORDS : { [key:string] : DART_KEYWORD_KIND } = {
```
mprobst: Do we actually need anything beyond the RESERVED keywords? If not, I'd just drop the BUILTIN and LIMITED stuff.

chalin: IMHO, a proper solution to #113 would be to:
* throw an error if an identifier is a RESERVED keyword and
* issue a warning otherwise.
Dart docs say that BUILTIN ids should be avoided but are ok to use.
Hence the need for the distinction in keywords.

lib/main.ts
```diff
@@ -5,6 +5,8 @@ require('source-map-support').install();
 import ts = require("typescript");
 import fs = require("fs");
 
+export enum DART_KEYWORD_KIND { RESERVED = 0, BUILTIN = 1, LIMITED = 2 };
```
mprobst: Should this enum be closer to its use site? I think you can define it within the class body, can't you?

lib/main.ts
mprobst: Sounds fine to me.

lib/main.ts
```diff
@@ -497,19 +499,30 @@ export class Transpiler {
...
-       ' set static super switch sync this throw true try typedef var void while with yield')
-          .split(/ /);
+  // Dart keyword list copied from
+  // https://www.dartlang.org/docs/dart-up-and-running/ch02.html#keywords
+  // Note the different kinds of keywords defined by the enum; key values
+  // match the footnote numbers given in the docs, or 0 if none given.
+  static DART_KEYWORDS : { [key:string] : DART_KEYWORD_KIND } = {
+    'abstract':1,	'continue':0,	'false':0,	'new':0,	'this':0,
+    'as':1,	'default':0,	'final':0,	'null':0,	'throw':0,
+    'assert':0,	'deferred':1,	'finally':0,	'operator':1,	'true':0,
```
mprobst: As these are enum values, they really should use the enum. It might be easier to actually have two sets and just do a lookup in them, rather than having the enum.

Also please mind the formatting here. I had it as a string with split as I thought that was more readable than a key per line, YMMV.

lib/main.ts
chalin: I had tried that earlier. It doesn't work.

lib/main.ts
chalin: Ok, I think I will revert to keywords lists then.

mprobst:
Sorry for the delay, I didn't get that you already did the fix :-)

chalin:
Np, I figured that you might have been busy. Will ping you next time via a comment.

PR #116, opened 2015-04-02
Disable filtering out CONST, IMPLEMENTS, ABSTRACT annotations.

mprobst:
It turns out that dartanalyze will complain about unused imports if
these annotations are no longer present. Without real type information
it's not possible to detect those cases reliably, so it's easier for now
to just include the annotations in the output.

Long term, we might either want to ignore those warnings or use the type
information from TS on what's actually used or isn't.

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -166,9 +166,9 @@ export class Transpiler {
     decorators.forEach((d) => {
       if (d.expression.kind !== ts.SyntaxKind.CallExpression) return;
       var funcExpr = <ts.CallExpression>d.expression;
-      if (this.ident(funcExpr.expression) !== 'IMPLEMENTS') return;
+      if (Transpiler.ident(funcExpr.expression) !== 'IMPLEMENTS') return;
       funcExpr.arguments.forEach((a) => {
-        var interf = this.ident(a);
+        var interf = Transpiler.ident(a);
         if (!interf) this.reportError(a, '@IMPLEMENTS only supports literal identifiers');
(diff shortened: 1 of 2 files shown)


lib/main.ts
```diff
@@ -305,6 +312,22 @@ export class Transpiler {
...
+
+  isEmptyImport(n: ts.ImportDeclaration): boolean {
+    var bindings = n.importClause.namedBindings;
+    if (bindings.kind != ts.SyntaxKind.NamedImports) return false;
+    return (<ts.NamedImports> bindings).elements.every((e) => {
+      return Transpiler.IGNORED_ANNOTATIONS.hasOwnProperty(this.ident(e.name));
+    });
+  }
+
+  filterImports(ns: ts.ImportOrExportSpecifier[]) {
```
alexeagle: it's a little scary to have this unused code, which may rot before we get around to re-enabling the feature.
How about making the predicate a static field?
```static isIgnoredAnnotation = (e) => {...}```
 then it could be used at line 321,
```elements.every(isIgnoredAnnotation)```
 and this filter method could be inlined below?

alexeagle:
Otherwise LGTM

lib/main.ts
mprobst: That's much nicer indeed. Doesn't even have to be a property, works simply as a static function (could also work as a top level function).

The dead code is a bit annoying indeed. Let's see that we get to TS soon. Worst case we'd just have to rewrite these couple of functions, which is the same as just deleting it...

PR #118, opened 2015-04-02
Re-format the source using clang-format.

mprobst:
Avoid an issue with multi-line strings by adjusting the code for
TemplateTail strings.


mprobst:
This seems a bit as if we should consider having the formatter fail the build, it's too easy to miss this.

alexeagle:
yes, let's fix this up today. It would be nice to start running clang-format in the angular repo (probably only on a few files to start)

PR #120, opened 2015-04-02
temporary: only read .js files, to work around lack of new broccoli filt...

alexeagle:
...ering

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -28,7 +28,7 @@ export class Transpiler {
     this.relativeFileName = relativeFileName;
     return program.getSourceFiles()
         .filter((sourceFile: ts.SourceFile) => !sourceFile.fileName.match(/\.d\.ts$/) &&
-                                               !!sourceFile.fileName.match(/\.[jt]s$/))
+                                               !!sourceFile.fileName.match(/\.[j]s$/))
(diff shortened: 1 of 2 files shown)


lib/main.ts
```diff
@@ -28,7 +28,7 @@ export class Transpiler {
     this.relativeFileName = relativeFileName;
     return program.getSourceFiles()
         .filter((sourceFile: ts.SourceFile) => !sourceFile.fileName.match(/\.d\.ts$/) &&
-                                               !!sourceFile.fileName.match(/\.[jt]s$/))
+                                               !!sourceFile.fileName.match(/\.[j]s$/))
```
mprobst: Maybe just change it to \.js$ then?

mprobst:
Please fix the e2e test, otherwise LGTM for the time being.

alexeagle:
broke on travis because of the version of typescript. It doesn't read .js files from the input.
Not sure what to do about it yet. I think it works with https://github.com/Microsoft/TypeScript/commit/b276c36a7bf54d68228a1f5a39f0c031929317ff but we don't want to be on that branch.

alexeagle:
dropping this since ts2dart should no longer read files we don't ask for

PR #121, opened 2015-04-02
Force TypeScript to transpile files one at a time

mprobst:
Cuts Angular's transpilation time from ~80 secs to 2.4 secs on my machine. x10 moon shots, baby!

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -18,7 +18,7 @@ export class Transpiler {
 
   constructor(private failFast: boolean = false, private generateLibraryName: boolean = false) {}
 
-  options: ts.CompilerOptions = {
+  static OPTIONS: ts.CompilerOptions = {
     target: ts.ScriptTarget.ES6,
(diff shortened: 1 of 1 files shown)


mprobst:
@alexeagle PTAL, this should be good to go in.

I'd need to mock "fs" to properly test this, or create a /tmp filesystem, but the code is trivial enough to not warrant that, I think.

lib/main.ts
```diff
@@ -33,15 +33,39 @@ export class Transpiler {
         .join('\n');
   }
 
+  createCompilerHost(files: string[]): ts.CompilerHost {
+    var fileMap: {[s: string]: boolean} = {};
```
alexeagle: isn't the object key always a string?

alexeagle:
LGTM

lib/main.ts
mprobst: Yes, but TS wants you to spell it out, for some reason.

PR #123, opened 2015-04-07
Optionally generate source maps.

mprobst:
This doesn't quite work yet - for some reason Chrome doesn't pick up the source map.

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -1,22 +1,32 @@
-/// <reference path="../typings/node/node.d.ts" />
+/// <reference path='../typings/node/node.d.ts' />
+/// <reference path='../typings/source-map/source-map.d.ts' />
 // Use HEAD version of typescript, installed by npm
-/// <reference path="../node_modules/typescript/bin/typescript.d.ts" />
+/// <reference path='../node_modules/typescript/bin/typescript.d.ts' />
 require('source-map-support').install();
-import ts = require("typescript");
-import fs = require("fs");
+import fs = require('fs');
+import SourceMap = require('source-map');
+import ts = require('typescript');
 
 export type ClassLike = ts.ClassDeclaration | ts.InterfaceDeclaration;
 
+export interface TranspilerOptions {
+  failFast?: boolean;
+  generateLibraryName?: boolean;
+  generateSourceMap?: boolean;
+}
+
 export class Transpiler {
-  result: string = '';
+  private output: Output;
+
+  private relativeFileName: string;
+  private currentFile: ts.SourceFile;
+
   // Comments attach to all following AST nodes before the next 'physical' token. Track the earliest
   // offset to avoid printing comments multiple times.
   lastCommentIdx: number = -1;
-  currentFile: ts.SourceFile;
-  relativeFileName: string;
   errors: string[] = [];
 
-  constructor(private failFast: boolean = false, private generateLibraryName: boolean = false) {}
+  constructor(private options: TranspilerOptions = {}) {}
 
   static OPTIONS: ts.CompilerOptions = {
     target: ts.ScriptTarget.ES6,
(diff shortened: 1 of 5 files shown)


mprobst:
@alexeagle PTAL, I think this is good.

lib/main.ts
```diff
@@ -1,22 +1,45 @@
...
   lastCommentIdx: number = -1;
-  currentFile: ts.SourceFile;
-  relativeFileName: string;
+
+  failFast: boolean;
   errors: string[] = [];
 
-  constructor(private failFast: boolean = false, private generateLibraryName: boolean = false) {}
+  constructor(opts: TranspilerOptions = {}) {
+    this.failFast = !!opts.failFast;
```
alexeagle: why do we store individual fields in the class, rather than a single TranspilerOptions reference?

lib/main.ts
```diff
@@ -26,11 +49,18 @@ export class Transpiler {
...
   translateProgram(program: ts.Program, relativeFileName: string): string {
     this.relativeFileName = relativeFileName;
-    return program.getSourceFiles()
-        .filter((sourceFile: ts.SourceFile) => !sourceFile.fileName.match(/\.d\.ts$/) &&
-                                               !!sourceFile.fileName.match(/\.[jt]s$/))
-        .map((f) => this.translate(f))
-        .join('\n');
+    var src = program.getSourceFiles()
+                  .filter((sourceFile: ts.SourceFile) => !sourceFile.fileName.match(/\.d\.ts$/) &&
+                                                         !!sourceFile.fileName.match(/\.[jt]s$/))
```
alexeagle: is this really clang-formatted?

lib/main.ts
```diff
@@ -26,11 +49,18 @@ export class Transpiler {
...
+                  .map((f) => this.translate(f))
+                  .join('\n');
+    if (this.sourceMap) src += this.generateSourceMapComment();
+    return src;
+  }
+
+  generateSourceMapComment() {
+    var base64map = new Buffer(JSON.stringify(this.sourceMap)).toString('base64');
+    return '\n\n//# sourceMappingURL=data:application/json;base64,' + base64map;
```
alexeagle: nit: template string rather than concatenation?

lib/main.ts
```diff
@@ -80,19 +110,51 @@ export class Transpiler {
...
       var e = new Error(this.errors.join('\n'));
       e.name = 'TS2DartError';
       throw e;
     }
+
     return this.result;
   }
 
+  enterNode(n: ts.Node) {
+    if (!this.sourceMap) return; // source maps disabled.
```
alexeagle: if this method is only for sourcemap generation, maybe it should have a different name?

lib/main.ts
```diff
@@ -80,19 +110,51 @@ export class Transpiler {
...
+    var file = n.getSourceFile() || this.currentFile;
+    var start = n.getStart(file);
+    var pos = file.getLineAndCharacterOfPosition(start);
+
+    var mapping: SourceMap.Mapping = {
+      original: {line: pos.line + 1, column: pos.character},
```
alexeagle: doesn't the character need a +1 also?

lib/main.ts
mprobst: This part is, it aligns the boolean clauses with the start of expression at `=>`. The source in createCompilerHost below sadly causes clang-format it to completely drop the ball. I'll investigate.

lib/main.ts
```diff
@@ -80,19 +110,51 @@ export class Transpiler {
...
+      generated: {line: this.line, column: this.column},
```
alexeagle: this means the source map only references the unformatted code, huh? maybe that's what you were trying to tell me yesterday :)

lib/main.ts
mprobst: No strong reason - I thought it was clearer to pull out the fields and coerce them into booleans here. I can change if you feel strongly about it.

lib/main.ts
```diff
@@ -773,7 +836,7 @@ export class Transpiler {
         this.emit(`'''${this.escapeTextForTemplateString(node)}'''`);
         break;
       case ts.SyntaxKind.TemplateMiddle:
-        this.result += this.escapeTextForTemplateString(node);
+        this.emitNoSpace(this.escapeTextForTemplateString(node));
```
alexeagle: these changes hint at a refactoring: maybe the result should be a private field in another class which represents the generated code. that class could encapsulate the responsibility of tracking the line and column

lib/main.ts
mprobst: YMMV, but for simple appending to a string I find this more readable.

alexeagle:
nice, LGTM
I guess you resolved the conflict between generating the sourcemap for dart vs. our code having sourcemap back to TS at runtime?

lib/main.ts
mprobst: Good point, done.

lib/main.ts
mprobst: I've experimentally established that it doesn't. It's odd, but until we get bug reports I'm willing to go with this :-)

lib/main.ts
mprobst: Yes :-)

lib/main.ts
mprobst: Done.

mprobst:
I've sent https://github.com/evanw/node-source-map-support/pull/48 for the source map issue.

PR #124, opened 2015-04-07
Target ES6.

mprobst:
This fixes the build as TSC started compiling about `let` statements in
ES5 targeting code. `typescript.d.ts` itself contains `let` statements,
so we have to move to ES6.

Also explicitly pass in the TypeScript dependency to make sure our
dependency versions are half-way sane.


alexeagle:
LGTM

mprobst:
This has been pushed as d9852b.

PR #126, opened 2015-04-12
Refactor all the things

mprobst:
The important part of this huge (sorry!) refactoring is `base.ts`.

This changes `Transpiler.visit` to dispatch to a sequence of transpiler steps that each implement one aspect of transpilation, by calling `visitNode` on the step. If a `TranspilerStep`'s `visitNode` method returns true, the node is considered to be handled, otherwise further steps can handle it. The `TranspilerStep` base class (in `base.ts`) provides basic support methods (`visitEach`, `emit` etc) and dispatches back through to `Transpiler.visit`.

`Transpiler` itself retains functionality for handling transpilation in general, source maps, files, errors, and output. It might make sense to pull it apart further. For now, this cleans up our "giant file syndrome".


mprobst:
Note that the ugly shim methods in base.ts dispatching back to transpiler go away in a later commit, those are only temporary.

alexeagle:
great!

mprobst:
Pushed as 836ac94.

mprobst:
All comments applied, except for the closing quote to fix the highlighter.

PR #127, opened 2015-04-16
Special case types ending in *Annotation.

mprobst:
This allows using decorators in TypeScript code and transpiling them to
annotations in Dart. It's still required to put @CONST annotations on the
annotation classes themselves.

### DIFF
--- a/test/decorator_test.ts
+++ b/test/decorator_test.ts
@@ -52,4 +52,15 @@ describe('decorators', () => {
     expectTranslate('@IMPLEMENTS(Z) class X implements Y {}')
         .to.equal(' @ IMPLEMENTS ( Z ) class X implements Y , Z { }');
   });
+
+});
+
+describe('annotation/decorator hack', () => {
+  it('should strip "Annotation" from type names', () => {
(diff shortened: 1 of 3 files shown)


test/decorator_test.ts
```diff
@@ -52,4 +52,15 @@ describe('decorators', () => {
...
         .to.equal(' @ IMPLEMENTS ( Z ) class X implements Y , Z { }');
   });
+
+});
+
+describe('annotation/decorator hack', () => {
+  it('should strip "Annotation" from type names', () => {
+    expectTranslate('@CONST class FooAnnotation {}').to.equal(' @ CONST const class Foo { }');
+  });
+  it('should strip "Annotation" from imports', () => {
```
rkirov: This reminds me that this feature can lead to some other subtle bugs when used the type is used for example,

```
class FooAnnotation extends BarAnnotation
```
but we can handle them as they arise.

rkirov:
LGTM.

test/decorator_test.ts
mprobst: Good observation, though outside of the parts that I'm overlooking - and I'm sure there are some - we should with this always strip "Annotation" from all type names, including super types. So This Should Not Happen (tm) (famous last words).

Issue #128, opened 2015-04-16
Don't install a node dependency in preinstall script


alexeagle:
The preinstall in package.json installs a version of tsd.
When angular tries to checksum the shrinkwrapped dependencies, it gets confused because the tsd shows up under node_modules even though it wasn't declared, so it always thinks the node_modules are dirty and busts the cache.
Maybe we can make tsd a devDependency and still get the preinstall to work somehow? Maybe it's enough to put it in both places.

alexeagle:
handled in 8ef71a190ee8927f6c03e38c3439f82a4fba1e67

PR #129, opened 2015-04-16
Move tsd to a regular dependency

alexeagle:


alexeagle:
Fixes #128

alexeagle:
(note, I'm surprised that the preinstall script succeeded - I expected it to complain that tsd had not yet been installed. but Travis passes. any idea why?)

mprobst:
LGTM

Wouldn't the preinstall succeed because it explicitly installs tsd?

alexeagle:
Yeah but preinstall before install right?

mprobst:
Yeah but this just called "npm install" in the preinstall, so that's how that worked? At least that'd be my theory.

alexeagle:
@IgorMinar : merged in 5af57c39cfaa46f2f89305310ce5c65628c139a6

PR #130, opened 2015-04-22
fix(annotations): adds two failing tests for Annotation renaming.

rkirov:


mprobst:
As discussed, this is obsolete, we'll use file names to distinguish the decorators vs annotations.

PR #134, opened 2015-04-28
fix(types): Handle instanceof identifiers.

mprobst:
@rkirov: we still need to handle type name references in instanceof expressions.


rkirov:
LGTM

PR #139, opened 2015-05-01
Use a forked version of node-source-map-support.

mprobst:
The main repository hasn't integrated our bug fix yet, giving bad stack traces
for files that contain more than one sourceMappingURL comment.


alexeagle:
LGTM

PR #144, opened 2015-05-04
feat(ts2dart): support property parameter shorthand.

### LINKED ISSUE #15: Support parameter properties
As per TypeScript [Parameter Properties](http://www.typescriptlang.org/Handbook#classes). Example:

AtScript:
```
class Foo {
  constructor(private _a:int);
}
```

Dart:
```
class Foo {
  int _a;
  Foo(this._a);
}
```

alexeagle:
This lets us declare properties in the constructor parameters in TypeScript code,
and emits explicit property declarations and the "this.prop" constructor param
which is the equivalent Dart shorthand.

Fixes #15

### DIFF
--- a/lib/declaration.ts
+++ b/lib/declaration.ts
@@ -87,27 +87,13 @@ class DeclarationTranspiler extends base.TranspilerStep {
           this.emit('const');
         }
         this.visit(className);
-        this.visitParameters(ctorDecl);
+        this.visitParameters(ctorDecl.parameters);
         this.visit(ctorDecl.body);
         break;
       case ts.SyntaxKind.PropertyDeclaration:
         var propertyDecl = <ts.PropertyDeclaration>node;
         this.visitDeclarationMetadata(propertyDecl);
-        var hasConstCtor = this.isConst(<base.ClassLike>propertyDecl.parent);
-        if (hasConstCtor) {
-          this.emit('final');
-        }
-        if (propertyDecl.type) {
-          this.visit(propertyDecl.type);
-        } else if (!hasConstCtor) {
-          this.emit('var');
-        }
-        this.visit(propertyDecl.name);
-        if (propertyDecl.initializer) {
-          this.emit('=');
-          this.visit(propertyDecl.initializer);
-        }
-        this.emit(';');
+        this.visitProperty(propertyDecl, <base.ClassLike>node.parent);
         break;
       case ts.SyntaxKind.SemicolonClassElement:
         // No-op, don't emit useless declarations.
@@ -138,7 +124,7 @@ class DeclarationTranspiler extends base.TranspilerStep {
         if (arrowFunc.body.kind == ts.SyntaxKind.Block) {
           this.visitFunctionLike(arrowFunc);
         } else {
-          this.visitParameters(arrowFunc);
+          this.visitParameters(arrowFunc.parameters);
           this.emit('=>');
           this.visit(arrowFunc.body);
         }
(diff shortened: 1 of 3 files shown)


lib/declaration.ts
```diff
@@ -82,12 +82,27 @@ class DeclarationTranspiler extends base.TranspilerStep {
           }
         }
         if (!className) this.reportError(ctorDecl, 'cannot find outer class node');
+        
+        let normalizedParameters: ts.ParameterDeclaration[] = [];
+        // Synthesize explicit properties for 'property parameters'
+        ctorDecl.parameters.forEach((param) => {
+            if (this.hasFlag(param.modifiers, ts.NodeFlags.Public) ||
```
mprobst: indent

lib/declaration.ts
```diff
@@ -246,7 +261,7 @@ class DeclarationTranspiler extends base.TranspilerStep {
     }
   }
 
-  private visitParameters(fn: ts.FunctionLikeDeclaration) {
+  private visitParameters(fn: {parameters: ts.ParameterDeclaration[]}) {
```
mprobst: Maybe just take parameters?

lib/declaration.ts
```diff
@@ -272,6 +287,26 @@ class DeclarationTranspiler extends base.TranspilerStep {
...
+  
+  /**
+   * Dart allows a shorthand for assigning to properties from the constructor, which is
+   * simply calling the parameter `this.name` where name is a property to assign.
+   * We must create these AST nodes to insert into the emit stream, since the original
+   * ParameterDeclaration gets emitted as the property declaration.
+   */
+  private buildSyntheticThisParam(param: ts.ParameterDeclaration): ts.ParameterDeclaration {
+    // The parse tree is immutable, so we build a copy by re-parsing from source.
+    let source = ts.createSourceFile("", `function f(${param.name.getText()}) {}`, ts2dart.OPTIONS.target, true);
```
mprobst: single quotes

mprobst:
Also can you add an integration test? My understanding was that this is valid TS:

```
class X { constructor(private y) {} }
```

But it'd have to translate to this Dart code:

```
class X {
  y;
  X(this.y) {}
}
```

I.e. in Dart you have to declare the field, but can then more easily initialize it. This code doesn't have the code to declare the field, does it?

alexeagle:
Added the integration test, and fixed up some of the AST re-write we talked about.
This snapshot still has the synthetic node to let us share the logic for writing out a property declaration, but I suspect we could remove that too. Will look again when I get to my desk.

lib/declaration.ts
```diff
@@ -273,6 +268,30 @@ class DeclarationTranspiler extends base.TranspilerStep {
...
   }
 
+  /**
+   * Visit a property declaration.
+   * In the special case of property parameters in a constructor, we also allow a parameter to be
+   * emitted as a property.
+   */
+  private visitProperty(propertyDecl: ts.PropertyDeclaration | ts.ParameterDeclaration,
+                        containingClass: base.ClassLike) {
+    var hasConstCtor = this.isConst(containingClass);
```
mprobst: inline?

test/e2e/lib.ts
```diff
@@ -1,15 +1,13 @@
 @CONST
 class MyClass {
-  private _field: string;
-
-  constructor(someVal: string) { this._field = someVal; }
+  constructor(private someVal: string) {}
```
mprobst: please make sure we have some code in e2e that uses a field

mprobst:
Looks very nice indeed!

test/e2e/lib.ts
alexeagle: fixed

alexeagle:
merged as d12a7b1131d64ae95d48f9248ec10d44d2db1314

PR #149, opened 2015-05-06
fix(build): typescript now a peerDependency

### LINKED ISSUE #148: conflict when user has different version of typescript
https://github.com/angular/angular/pull/1710
failed because of
```
TS2DartError: angular2/angular2.js:1:1: Unsupported node type SourceFile: export * from './change_detection';
export * from './core';
export * from './annotations';
export * from './directives';
export * from './forms';
export {Observable, EventEmitter} from 'angular2/src/facade/async';
export * from 'angular2/src/render/api';
export {DirectDomRenderer} from 'angular2/src/render/dom/direct_dom_renderer';
```

our theory is that this is due to version mismatch: the SourceFile enum is different between the version used by the caller of ts2dart and the one ts2dart depends on.

Probably should try peerDependency so the user of ts2dart gives us the typescript version to use.

alexeagle:
Fixes #148

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,12 +9,13 @@
   "dependencies": {
     "fs-extra": "^0.18.0",
     "source-map": "^0.4.2",
-    "source-map-support": "git://github.com/mprobst/node-source-map-support",
-    "typescript": "git://github.com/mhegazy/TypeScript.git#v1.5-beta"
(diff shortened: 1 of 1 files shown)


alexeagle:
merged as c9568754e0b342a2441d81493f47eb350f16b731

PR #151, opened 2015-05-08
feat(const): Translate CONST_EXPR to const expressions.

mprobst:
This allows to construct compile time literal (const) expressions in Dart, e.g.

```
const bools = CONST_EXPR([true, false, FileNotFound]);
```

### DIFF
--- a/lib/call.ts
+++ b/lib/call.ts
@@ -40,7 +40,14 @@ class CallTranspiler extends base.TranspilerStep {
   }
 
   private visitCall(c: ts.CallExpression) {
-    this.visit(c.expression);
+    var fnName = base.ident(c.expression);
+    if (fnName === 'CONST_EXPR') {
+      // The special function CONST_EXPR translates to Dart's const expressions.
+      this.emit('const');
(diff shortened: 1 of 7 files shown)


mprobst:
Allows progress on https://github.com/angular/angular/pull/1524.

@tbosch @pkozlowski-opensource FYI.

pkozlowski-opensource:
@mprobst awesome. Shell we have the `CONST_EXPR` defined on the JS/TS side somewhere as well to have one code working correctly for both JS and Dart?

mprobst:
@pkozlowski-opensource yes, the idea is that the TS code would have a global function in `lang.ts`.

Actually let me make sure we'll filter a `CONST_EXPR` import from the Dart output to make this a bit nicer.

mprobst:
This also fixes https://github.com/angular/ts2dart/issues/13 now.

tbosch:
+1!

mprobst:
@alexeagle I'm off for the day, please merge & release if you think this is fine, or take over the change if you think it isn't.

alexeagle:
Looks great, merging now.

PR #153, opened 2015-05-08
Preserve dart stack traces

alexeagle:
For https://github.com/angular/angular/issues/1407

### DIFF
--- a/test/statement_test.ts
+++ b/test/statement_test.ts
@@ -47,4 +47,7 @@ describe('statements', () => {
     expectTranslate('try {} catch (e) { console.log(e, e.stack); }')
         .to.equal(' try { } catch ( e , e_stack ) { console . log ( e , e_stack ) ; }');
   });
+  it('rewrites rethrow to preserve stack trace', () => {
+    expectTranslate('try {} catch (ex) { throw ex; }').to.equal(' try { } catch ( ex , ex_stack ) { rethrow ; }')
+  });
 });
--- a/test/statement_test.ts
+++ b/test/statement_test.ts
@@ -30,9 +30,9 @@ describe('statements', () => {
   });
   it('translates try/catch', () => {
     expectTranslate('try {} catch(e) {} finally {}')
-        .to.equal(' try { } catch ( e ) { } finally { }');
+        .to.equal(' try { } catch ( e , e_stack ) { } finally { }');
     expectTranslate('try {} catch(e: MyException) {}')
-        .to.equal(' try { } on MyException catch ( e ) { }');
+        .to.equal(' try { } on MyException catch ( e , e_stack ) { }');
   });
   it('translates throw', () => {
     expectTranslate('throw new Error("oops")').to.equal(' throw new Error ( "oops" ) ;');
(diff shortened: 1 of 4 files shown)


test/statement_test.ts
```diff
@@ -30,9 +30,9 @@ describe('statements', () => {
   });
   it('translates try/catch', () => {
     expectTranslate('try {} catch(e) {} finally {}')
-        .to.equal(' try { } catch ( e ) { } finally { }');
+        .to.equal(' try { } catch ( e , e_stack ) { } finally { }');
```
mhevery: ideally we should only get the stack if we need it, as it is expensive.

yjbanov: Dart, like JS, has `Error`, which has a getter `stackTrace`. We could, instead of `catch(e, s)`, do this:

```dart
} catch(Error e) {
  print(e.stackTrace);
}
```

All Dart's own standard errors are subclasses of `Error` already, and so is our `BaseException`. So it should work just fine.

alexeagle: How 'ideally' is that? It's challenging to look ahead in that way.

alexeagle: Yegor, if I understand correctly you're saying no change is needed in
ts2dart because we should already be catching Error in the original angular
code? What change needs to be made in that case?

mprobst: @mhevery if catching the stack is that expensive, surely the VM optimizes for the case where the stack isn't actually used subsequently, and we can just generate stupid code that always captures the stack?

alexeagle: Or dart2js could detect the unused variable.

yjbanov: To be precise, what I'm proposing long-term is that we use the common syntax in both Dart and JS and access stack via a facade (because the field names are different; it's `err.stack` vs `err.stackTrace`). Example:

```dart
// facade/error.dart
class ErrorWrapper {
  static stackTrace(Error err) => err.stackTrace;
}
```

```typescript
// facade/error.ts
export class ErrorWrapper {
  static stackTrace(err:Error) {
    return err.stack;
  }
}
```

Then in all places in Angular 2 code we would:

```typescript
try {
  ...
} catch(err) {
  var st = ErrorWrapper.stackTrace(err);
}
```

Works in both languages.

alexeagle: I would be concerned that it's hard to enforce using the facade since
coding the TS idiom would only break error reporting.
Why not code in idiomatic TypeScript and transpile from your second example
to the first? We just need to understand the types (will have soon)

yjbanov: @mhevery, I believe it is only expensive when an error actually happens. During normal operation, errors should be very rare, so it should be ok to ask for stack trace for every error. @floitschG, could you please confirm this assumption?

floitschG:
Dart2js works with JS Errors. That is, if you throw an error from within Dart, the error is wrapped into a JS error. For one this preserves the stacktrace, and it also makes it uniform with implicit errors.
When catching an error, dart2js builds the stacktrace local by instantiating the StackTrace class with the error as argument. During a toString() of this instance the `.stack` property of the JS error is extracted and returned.
This means that adding the stacktrace variable to the catch handler is cheap (an allocation of an instance with 2 fields).

Furthermore dart2js currently always extracts the stacktrace anyway. It should be easy to fix that in dart2js, but in the end it is not that expensive...
(The reason it always happens is that the extraction of the stacktrace is seen as having a side-effect, and is therefore not eliminated even though the stacktrace local is never used).

yjbanov:
@floitschG, so it sounds to me that there is no difference at all if the code is not actually throwing anything. During normal operation, the application is not throwing many errors. We only catch errors for debugging.

floitschG:
If the code is not throwing then it won't matter if you catch the stacktrace or not.
Even if you throw, dart2js has to include the stacktrace because it can't know if some error-handler will need it. For that it would need to dynamically look for the next error-handler (which the Dart VM can do).
So in short: adding the stacktrace parameter won't cost you.

yjbanov:
Excellent! Thanks, Florian. Then this PR is good to go.

alexeagle:
merged.

PR #154, opened 2015-05-10
fix(const): fix processing of CONST_EXPR

pkozlowski-opensource:
With the current implementation `const numbers = CONST_EXPR([1, 2, 3]);` would result in `const numbers = const([1, 2, 3]);` which doesn't seem to be valid Dart code as it results in:

> error: type name expected

I think we should be simply generating `const numbers = const [1, 2, 3];`

### DIFF
--- a/lib/call.ts
+++ b/lib/call.ts
@@ -41,18 +41,22 @@ class CallTranspiler extends base.TranspilerStep {
 
   private visitCall(c: ts.CallExpression) {
     var fnName = base.ident(c.expression);
-    if (fnName === 'CONST_EXPR') {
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@alexeagle @mprobst could you guys please have a look and merge + release a new version if this sounds about right to you?

alexeagle:
Opt: would be nice to add assertion in the e2e test so we know the unit test is correct too.

LGTM, but I am in Yosemite park and mostly offline so you'll have to merge and release.

mprobst:
Fixed in #155.

PR #155, opened 2015-05-10
fix(const): Fix const expressions transpilation.

mprobst:


mprobst:
@pkozlowski-opensource FYI, taking over your pull request if you don't mind.

This is essentially the same fix as #154 but uses an early exist.

alexeagle:
Nice thanks for the e2e test. Looks like we need to add formatting enforcement?

mprobst:
Yes, we should.

Alex Eagle <notifications@github.com> schrieb am So., 10. Mai 2015 18:05:

PR #156, opened 2015-05-13
chore(types): Switch to --noImplicitAny mode.

mprobst:
This disallows expressions to have implicit "any" types, i.e. expressions must
have their type inferred or must be typed. We should try this mode out a bit
and see if it's useful or problematic, as a test bed for Angular and other
places.

### DIFF
--- a/lib/declaration.ts
+++ b/lib/declaration.ts
@@ -74,7 +74,7 @@ class DeclarationTranspiler extends base.TranspilerStep {
       case ts.SyntaxKind.Constructor:
         var ctorDecl = <ts.ConstructorDeclaration>node;
         // Find containing class name.
-        var className;
+        var className: ts.Identifier;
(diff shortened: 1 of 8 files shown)


lib/declaration.ts
```diff
@@ -244,7 +244,7 @@ class DeclarationTranspiler extends base.TranspilerStep {
 
   private visitParameters(parameters: ts.ParameterDeclaration[]) {
     this.emit('(');
-    let firstInitParamIdx;
+    let firstInitParamIdx: number;
```
alexeagle: Maybe initialize instead of declaring type, then omit loop initializer

alexeagle:
Sounds like a great idea to me.

lib/declaration.ts
mprobst: Done.

PR #161, opened 2015-05-18
chore(build): do not install and run tsd on every install.

mprobst:
Install scripts are run by every user on every install, but ts2dart users do not actually need tsd or the type definitions, as we ship compiled JavaScript. This should speed up downstream project builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,8 +26,7 @@
     "which": "^1.0.9"
   },
   "scripts": {
-    "preinstall": "npm install tsd@^0.5.7 && tsd reinstall",
-    "prepublish": "gulp compile",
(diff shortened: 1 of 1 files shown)


alexeagle:
I am scared by changes here because of the large number of ways it can be broken. (eg https://github.com/angular/ts2dart/issues/138 )
But this change *looks* reasonable, and making angular npm install faster is worth the risk of breaking the shrinkwrap again, and I've gotten exceedingly good at fixing it.

mprobst:
Ack, but this is actually removing `tsd` from the `npm install` story of `ts2dart`, so I think this reduces risk, if anything.

PR #163, opened 2015-05-19
feat: Add a tsconfig.json.

mprobst:
This is a minimal configuration to get rid of warnings about not specifying a
module system in SublimeText. It disables emitting, as this is handled by the
actual build running via gulp.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,7 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "noImplicitAny": true,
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "noImplicitAny": true,
+    "noEmit": true,
```
alexeagle: @tbosch this is a nice idea, we don't want the angular one to emit either

PR #166, opened 2015-05-19
fix(bug): dart abstract methods don't use abstract keyword

### LINKED ISSUE #165: Interfaces should not create abstract methods
An interface in TS is transpiled right now as follows:

Typescript:
```
interface Test {
  doIt();
}
```

Dart:
```
abstract class Test {
  abstract doIt();
}
```

However, Dart complains about this, and removing the `abstract` from the method `doIt()` seems to resolve this.

alexeagle:
fixes #165


alexeagle:
merged in bc906ab26209ec1be5e04efe32e9633be0a8e887

PR #170, opened 2015-05-20
added support for union types

vsavkin:


lib/type.ts
```diff
@@ -12,6 +12,12 @@ class TypeTranspiler extends base.TranspilerStep {
         // Dart doesn't support type literals.
         this.emit('dynamic');
         break;
+      case ts.SyntaxKind.UnionType:
+        this.emit('dynamic /*');
+        var t: any = node;
```
tbosch: Why is there `any`?

PR #171, opened 2015-05-20
fix(const): fixed const expressions for constructor

vsavkin:


lib/call.ts
```diff
@@ -19,11 +19,25 @@ class CallTranspiler extends base.TranspilerStep {
...
           this.emit('new');
         }
         this.visitCall(<ts.NewExpression>node);
         break;
+      case ts.SyntaxKind.ArrayLiteralExpression:
+        if (this.isInsideConstExpr(node)) {
+          this.emit('const');
+        }
+        return false;
+        break;
```
tbosch: Don't need the break here.

lib/call.ts
```diff
@@ -19,11 +19,25 @@ class CallTranspiler extends base.TranspilerStep {
...
+      case ts.SyntaxKind.ObjectLiteralExpression:
+        if (this.isInsideConstExpr(node)) {
+          this.emit('const');
+        }
+        return false;
+        break;
```
tbosch: Don't need break here.

PR #181, opened 2015-06-04
feat: façade converter handles various special cases for Dart.

mprobst:
- Bundles CONST_EXPR and FORWARD_REF into one place
- Special cases a bunch of Array methods, allowing more idiomatic TypeScript
  & Dart code
- Code structure should allow straight forward future additions.

### DIFF
--- a/lib/call.ts
+++ b/lib/call.ts
@@ -2,9 +2,10 @@
 import ts = require('typescript');
 import base = require('./base');
 import ts2dart = require('./main');
+import {FacadeConverter} from "./facade_converter";
 
-class CallTranspiler extends base.TranspilerStep {
-  constructor(tr: ts2dart.Transpiler) { super(tr); }
+class CallTranspiler extends base.TranspilerBase {
+  constructor(tr: ts2dart.Transpiler, private fc: FacadeConverter) { super(tr); }
 
   visitNode(node: ts.Node): boolean {
     switch (node.kind) {
@@ -26,21 +27,11 @@ class CallTranspiler extends base.TranspilerStep {
         }
         this.visitCall(<ts.NewExpression>node);
         break;
-      case ts.SyntaxKind.ArrayLiteralExpression:
-        if (this.isInsideConstExpr(node)) {
-          this.emit('const');
-        }
-        return false;
-      case ts.SyntaxKind.ObjectLiteralExpression:
-        if (this.isInsideConstExpr(node)) {
-          this.emit('const');
-        }
-        return false;
       case ts.SyntaxKind.CallExpression:
         var callExpr = <ts.CallExpression>node;
-        if (!this.maybeHandleSuperCall(callExpr)) {
-          this.visitCall(callExpr);
-        }
+        if (this.fc.maybeHandleCall(callExpr)) break;
+        if (this.maybeHandleSuperCall(callExpr)) break;
+        this.visitCall(callExpr);
         break;
       case ts.SyntaxKind.SuperKeyword:
         this.emit('super');
@@ -52,30 +43,6 @@ class CallTranspiler extends base.TranspilerStep {
   }
 
   private visitCall(c: ts.CallExpression) {
-    if (this.isConstCall(c)) {
-      // The special function CONST_EXPR translates to Dart's const expressions.
-      if (c.arguments.length !== 1) this.reportError(c, 'CONST_EXPR takes exactly one argument');
-      this.visitList(c.arguments);
-      return;
-    }
-
-    if (this.isForwardRef(c)) {
-      // The special function FORWARD_REF translated to an unwrapped value
-      if (c.arguments.length !== 1) {
-        this.reportError(c, 'FORWARD_REF takes exactly one argument');
-        return;
-      }
-
-      const callback = <ts.FunctionExpression>c.arguments[0];
-      if (callback.kind !== ts.SyntaxKind.ArrowFunction) {
-        this.reportError(c, 'FORWARD_REF takes only arrow functions');
-        return;
-      }
-
-      this.visit(callback.body);
-      return;
-    }
-
     this.visit(c.expression);
     this.emit('(');
     if (!this.handleNamedParamsCall(c)) {
--- a/lib/expression.ts
+++ b/lib/expression.ts
@@ -2,9 +2,10 @@
 import ts = require('typescript');
 import base = require('./base');
 import ts2dart = require('./main');
+import {FacadeConverter} from './facade_converter';
 
-class ExpressionTranspiler extends base.TranspilerStep {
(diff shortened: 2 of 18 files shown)


mprobst:
/CC @mhevery 

Partially handles angular/angular#2231.

lib/call.ts
```diff
@@ -26,21 +27,11 @@ class CallTranspiler extends base.TranspilerStep {
...
-        if (this.isInsideConstExpr(node)) {
-          this.emit('const');
-        }
-        return false;
       case ts.SyntaxKind.CallExpression:
         var callExpr = <ts.CallExpression>node;
-        if (!this.maybeHandleSuperCall(callExpr)) {
-          this.visitCall(callExpr);
-        }
+        if (this.fc.handleCall(callExpr)) break;
```
alexeagle: I like the maybe* naming convention, want to use it here too?

alexeagle: opt: could use boolean OR rather than break, eg.  `this.fc.handle || this.maybeHandleSuper || this.visitCall`

lib/expression.ts
```diff
@@ -2,9 +2,10 @@
 import ts = require('typescript');
 import base = require('./base');
 import ts2dart = require('./main');
```
alexeagle: these require statements should be ES6 module instead?

lib/facade_converter.ts
```diff
@@ -0,0 +1,150 @@
...
+    var qnSub = fileSubs[qn];
+    if (!qnSub) return false;
+
+    qnSub(c, context);
+    return true;
+  }
+
+  private subs: ts.Map<ts.Map<FacadeHandler>> = {
+    'lib': {
+      'Array.push': (c: ts.CallExpression, context: ts.Expression) => {
```
alexeagle: can type inference work here, so you could just `(c, context) => {}`
would make this structure more readable.

lib/facade_converter.ts
```diff
@@ -0,0 +1,150 @@
+/// <reference path='../node_modules/typescript/bin/typescript.d.ts' />
+import base = require('./base');
+import ts = require('typescript');
+import ts2dart = require('./main');
+
+type FacadeHandler = (c: ts.CallExpression, context: ts.Expression) => void;
+
+export class FacadeConverter extends base.TranspilerStep {
```
alexeagle: seems unintuitive that this is a base.TranspilerStep but it doesn't handle any nodes on its own, instead is just a helper. Is it really a subtype?

lib/facade_converter.ts
```diff
@@ -0,0 +1,150 @@
...
+    super(transpiler);
+    this.forbiddenNames = {};
+    for (var fileName in this.subs) {
+      Object.keys(this.subs[fileName])
+          .map((fnName) => fnName.substring(fnName.lastIndexOf('.') + 1))
+          .forEach((fnName) => this.forbiddenNames[fnName] = true);
+    }
+  }
+
+  setTypeChecker(tc: ts.TypeChecker) { this.tc = tc; }
```
alexeagle: still don't like ES6 setter syntax? :)

lib/main.ts
```diff
@@ -74,6 +86,9 @@ export class Transpiler {
     }
     var destinationRoot = destination || this.options.basePath || '';
     var program = ts.createProgram(fileNames, OPTIONS, host);
+    if (this.options.translateBuiltins) {
+      this.fc.setTypeChecker(program.getTypeChecker());
```
alexeagle: why not make the typeChecker a constructor param on FacadeConverter?

test/facade_converter_test.ts
```diff
@@ -0,0 +1,75 @@
...
+      new<K, V>(l: List<any>): Map<K, V>;
+      prototype: Map<any, any>;
+    };`;
+
+var langDeclarations = `
+  export declare function CONST_EXPR<T>(x: T): T;
+  export declare function FORWARD_REF<T>(x: T): T;
+  `;
+
+describe('collection façade', () => {
```
alexeagle: you had to put it somewhere...

test/facade_converter_test.ts
```diff
@@ -0,0 +1,75 @@
...
+        .to.equal(' const x = const { "one" : 1 } ;');
+  });
+
+  it('translates FORWARD_REF(() => T) to T', () => {
+    expectWithTypes('import {FORWARD_REF} from "angular2/src/facade/lang";\n' +
+                    'var x = FORWARD_REF(() => SomeType);')
+        .to.equal(' var x = SomeType ;');
+    chai.expect(() => expectWithTypes('import {FORWARD_REF} from "angular2/src/facade/lang";\n' +
+                                      'FORWARD_REF(1)'))
+        .to.throw(/only arrow functions/);
```
alexeagle: ooh that is very pretty syntax for less brittle exception assertion

lib/call.ts
mprobst: Done re maybe convention.
Regarding the ||, I think this is not much more code and reads a bit cleaner to me.

lib/expression.ts
mprobst: Ultimately, yes, but let's do that in another commit.

lib/facade_converter.ts
mprobst: As discussed, renamed this to TranspilerBase.

lib/facade_converter.ts
mprobst: Still don't like setters, ES6 or not ;-)

lib/main.ts
mprobst: Yeah, the transpiler steps are created before creating the program (they can be used multiple times over different programs), and I can only get the TypeChecker from the program, but not from the individual nodes we're visiting. The alternative would be passing down the TypeChecker with the program nodes, but that seems overkill.

test/test_support.ts
```diff
@@ -1,31 +1,57 @@
...
-      if (sourceName === fileName) {
-        return ts.createSourceFile(sourceName, contents, compilerOptions.target, true);
+      if (nameToContent.hasOwnProperty(sourceName)) {
+        return ts.createSourceFile(sourceName, nameToContent[sourceName], compilerOptions.target,
+                                   true);
       }
       if (sourceName === 'lib.d.ts') {
-        return ts.createSourceFile(sourceName, '', compilerOptions.target, true);
+        if (!libSourceFile) {
+          // Cache to avoid excessive test times.
```
alexeagle: potential for test isolation failures? are the cache contents immutable?

alexeagle:
LGTM when Travis says so

test/test_support.ts
mprobst: Yes, everything in tsc is immutable AFAIK.

PR #182, opened 2015-06-04
chore: move tsd to a devDependency for faster npm install times.

### LINKED ISSUE #159: tsd should be devDependency and not a dependency
Since tsd is now a `dependency` in package.json all projects that depend on ts2dart have to install it. tsd and it's dependencies are quite sizable so that's far from ideal.

Making tsd a devDependency resolves this issue because devDependencies are not part of the transitive closure of dependencies in a project that uses ts2dart.

Note: I know that this complicates the publish process, but maybe replacing that with a shell script or some task runner would be a better choice.

mprobst:
Fixes #159.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,6 @@
     "fs-extra": "^0.18.0",
     "source-map": "^0.4.2",
     "source-map-support": "^0.3.1",
-    "tsd": "^0.6.0",
     "typescript": "alexeagle/TypeScript#error_is_class"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


alexeagle:
I see Igor's comment about this, but I having it as a devDependency broke angular.
https://github.com/angular/ts2dart/issues/138
which I fixed by moving from devDependency to regular dependency.

I think our only fix here is to write a release script rather than rely on npm to perform the right things in the right order.

mprobst:
@alexeagle yeah, but that was when we ran `tsd reinstall` in `npm install` as a preinstall script, right? Now that we only run it on publish, we should be fine.

Issue #145, opened 2015-05-05
Comments are not retained on constructors


alexeagle:
This test code
```
      expectTranslate('/** A */ class A { /** ctor */ constructor() {}}')
          .to.equal(' /** A */ class A { /** ctor */ A() { } }')
```
fails with
```
AssertionError: expected ' /** A */ class A { A ( ) { } }' to equal ' /** A */ class A { /** ctor */ A() { } }'
```

PR #183, opened 2015-06-04
feat: test case for constructor comments.

mprobst:
These were always generated, but only if there is a line break before the
comment start, as otherwise TS considers the comment to belong to the pre-
ceding token.


mprobst:
Fixes #145.

alexeagle:
LGTM

PR #187, opened 2015-06-04
feat: ignore return types for arrow functions.

### LINKED ISSUE #179: Closure & return type
`(_, __): string => { throw new BaseException(`Failed to fetch url "${url}"`); });`
->
`String (_, __) { ... }` 

which is invalid in Dart

mprobst:
Dart does not support closures with return types.

Fixes #179.

### DIFF
--- a/lib/declaration.ts
+++ b/lib/declaration.ts
@@ -223,7 +223,16 @@ class DeclarationTranspiler extends base.TranspilerStep {
   }
 
   private visitFunctionLike(fn: ts.FunctionLikeDeclaration, accessor?: string) {
-    if (fn.type) this.visit(fn.type);
+    if (fn.type) {
+      if (fn.kind === ts.SyntaxKind.ArrowFunction) {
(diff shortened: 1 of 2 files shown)


lib/declaration.ts
```diff
@@ -223,7 +223,8 @@ class DeclarationTranspiler extends base.TranspilerStep {
   }
 
   private visitFunctionLike(fn: ts.FunctionLikeDeclaration, accessor?: string) {
-    if (fn.type) this.visit(fn.type);
+    // Type is silently dropped for arrow functions, not supported in Dart.
+    if (fn.type && fn.kind !== ts.SyntaxKind.ArrowFunction) this.visit(fn.type);
```
alexeagle: should emit the type inside a comment like we do in other places

mprobst: Done.

mprobst:
Build breakage is a flake.

alexeagle:
LGTM

We have flakes already?

mprobst:
It's not our flake, really.

PR #193, opened 2015-06-13
fix: properly handle regexp literals.

mprobst:
Translate to "new RegExp" calls with raw strings.
Handle .text and .exec in the façade.

@jelbourn


jelbourn:
LGTM insofar as I know anything about ts2dart or RegExp in dart.

PR #196, opened 2015-06-14
fix: normalize slashes to support windows.

### LINKED ISSUE #191: Normalize slashes in createCompilerHost to support windows
In windows default path separator is backslash. Typescript internally normalizes slashes and converts them to /, so when it queries the host for getSourceFile, the sourceName won't match with the any entry in fileMap and hence the files don't get transpiled.

piloopin:
Closes #191.


mprobst:
Fixed in 070ff40348e198e683b60376ea674e3eca132b0f, thanks for the contribution!

PR #198, opened 2015-06-15
feat: support regular expression flags, e.g. /abc/gi.

### LINKED ISSUE #195: RegExp literal
It seems like flags (g, i, m)  are not handled in RegExp literal at https://github.com/angular/ts2dart/blob/c34c27431026bde4013bbee9daa0228028c7b916/lib/literal.ts#L95

/cc @mprobst

mprobst:
Fixes #195.

PTAL @vicb.

### DIFF
--- a/lib/literal.ts
+++ b/lib/literal.ts
@@ -102,6 +102,10 @@ class LiteralTranspiler extends base.TranspilerBase {
         regExp = regExp.replace('\'', '\' + "\'" + r\'');  // handle nested quotes by concatenation.
         this.emitNoSpace(regExp);
         this.emitNoSpace('\'');
+        if (flags.indexOf('g') === -1) {
+          // Dart RegExps are always global, so JS regexps must use 'g' so that semantics match.
+          this.reportError(node, 'Regular Expressions must use the //g flag');
+        }
         if (flags.indexOf('m') !== -1) {
           this.emit(', multiline: true');
         }
--- a/lib/literal.ts
+++ b/lib/literal.ts
@@ -96,10 +96,18 @@ class LiteralTranspiler extends base.TranspilerBase {
         this.emit('new RegExp (');
         this.emit('r\'');
         var regExp = (<ts.LiteralExpression>node).text;
-        regExp = regExp.substring(1, regExp.length - 1);  // cut off /.../ chars.
+        var slashIdx = regExp.lastIndexOf('/');
+        var flags = regExp.substring(slashIdx + 1);
+        regExp = regExp.substring(1, slashIdx);  // cut off /.../ chars.
         regExp = regExp.replace('\'', '\' + "\'" + r\'');  // handle nested quotes by concatenation.
         this.emitNoSpace(regExp);
         this.emitNoSpace('\'');
+        if (flags.indexOf('m') !== -1) {
+          this.emit(', multiline: true');
+        }
+        if (flags.indexOf('i') !== -1) {
+          this.emit(', caseSensitive: false');
+        }
         this.emit(')');
         break;
       case ts.SyntaxKind.ThisKeyword:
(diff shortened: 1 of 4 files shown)


lib/literal.ts
```diff
@@ -96,10 +96,18 @@ class LiteralTranspiler extends base.TranspilerBase {
...
         this.emit('r\'');
         var regExp = (<ts.LiteralExpression>node).text;
-        regExp = regExp.substring(1, regExp.length - 1);  // cut off /.../ chars.
+        var slashIdx = regExp.lastIndexOf('/');
+        var flags = regExp.substring(slashIdx);
+        regExp = regExp.substring(1, slashIdx);  // cut off /.../ chars.
         regExp = regExp.replace('\'', '\' + "\'" + r\'');  // handle nested quotes by concatenation.
         this.emitNoSpace(regExp);
         this.emitNoSpace('\'');
+        if (flags.indexOf('g') !== -1) {
```
vicb: I think it should be `m`. `g` should have no effect in Dart.

lib/literal.ts
```diff
@@ -96,10 +96,18 @@ class LiteralTranspiler extends base.TranspilerBase {
         this.emit('new RegExp (');
         this.emit('r\'');
         var regExp = (<ts.LiteralExpression>node).text;
-        regExp = regExp.substring(1, regExp.length - 1);  // cut off /.../ chars.
+        var slashIdx = regExp.lastIndexOf('/');
+        var flags = regExp.substring(slashIdx);
```
vicb: +1 ?

vicb: `slashIdx + 1`

lib/literal.ts
mprobst: Fixed. I also added a change to enforce all regexps to have 'g'. PTAL.

lib/literal.ts
mprobst: Done. I noticed, but it doesn't matter, so I was lazy ;-)

lib/literal.ts
```diff
@@ -96,10 +96,22 @@ class LiteralTranspiler extends base.TranspilerBase {
...
         var regExp = (<ts.LiteralExpression>node).text;
-        regExp = regExp.substring(1, regExp.length - 1);  // cut off /.../ chars.
+        var slashIdx = regExp.lastIndexOf('/');
+        var flags = regExp.substring(slashIdx + 1);
+        regExp = regExp.substring(1, slashIdx);  // cut off /.../ chars.
         regExp = regExp.replace('\'', '\' + "\'" + r\'');  // handle nested quotes by concatenation.
         this.emitNoSpace(regExp);
         this.emitNoSpace('\'');
+        if (flags.indexOf('g') === -1) {
+          // Dart RegExps are always multiline, so JS regexps must use 'g' so that semantics match.
```
vicb: g is still not "multiline" but "global" :)

I'm unsure about this if branch, depends on what we do for `exec` ? I guess it good for now.

vicb:
@mprobst I don't think this cover the "other" syntaxes:

```
new RegExp('ab+c', 'i');
new RegExp(/ab+c/, 'i');
```

The 2nd one might especially be problematic

mprobst:
@vicb it's not intended to support, that's a pretty different problem. Do you think we need the explicit ctor for anything? Otherwise I'd say just don't do that ;-)

vicb:
The problem is that it would generate invalid Dart.

Is that complex to support ?

mprobst:
It won't be super hard, but while it adds the ability to have runtime generated regular expressions, I don't think it's a big use case so far, is it?

I'm generally fine with ts2dart spitting out code that's invalid Dart, as that'd be caught by the Dart analyzer. ts2dart does not aim to be a complete translator for all TypeScript to Dart, we're just translating the subset that's used and useful in AngularJS.

vicb:
Thanks for clarification.

Then LGTM after multiline has been changed to global (inline comment)

PR #199, opened 2015-06-15
fix: paths for Windows

mlaval:
Follow-up of #196 to make tests pass on Windows, use `path.isAbsolute` and fix an issue in the meantime (all platforms):
`chai.expect(getNormalizedPath(transpiler, '/a/b/c.js', '/e/f/g/h/i/j/k/l')).to.equal('/a/b/c.dart')`

[Review on Reviewable](https://reviewable.io/reviews/angular/ts2dart/199)

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -156,9 +156,15 @@ export class Transpiler {
 
   // Visible for testing.
   getOutputPath(filePath: string, destinationRoot: string): string {
-    var relative = this.getRelativeFileName(filePath);
-    var dartFile = relative.replace(/.(js|es6|ts)$/, '.dart');
-    return path.join(destinationRoot, dartFile);
+    var dartFile = filePath.replace(/.(js|es6|ts)$/, '.dart');
+    var base = this.options.basePath || '';
+    if (this.isAbsolute(filePath) && this.isAbsolute(destinationRoot) &&
+        filePath.indexOf(destinationRoot) !== 0 && (base === undefined || base === '')) {
+      return dartFile;
+    } else {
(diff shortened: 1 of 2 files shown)


mlaval:
`path.isAbsolute` doesn't exist in Node 0.10 ... worked around that.

lib/main.ts
```diff
@@ -156,9 +156,15 @@ export class Transpiler {
 
   // Visible for testing.
   getOutputPath(filePath: string, destinationRoot: string): string {
-    var relative = this.getRelativeFileName(filePath);
-    var dartFile = relative.replace(/.(js|es6|ts)$/, '.dart');
-    return path.join(destinationRoot, dartFile);
+    var dartFile = filePath.replace(/.(js|es6|ts)$/, '.dart');
```
mprobst: I think it's time to give up here and use some proper library call to make the path relative.

I don't know why we're not using `path.resolve(destinationRoot, relative)` here. Care to try?

mlaval: This part of the PR was in fact to solve a use case I found out during investigation:
`chai.expect(getNormalizedPath(transpiler, '/a/b/c.js', '/e/f/g/h/i/j/k/l')).to.equal('/a/b/c.dart');`

This is not a Windows thing, it was failing on all platforms.
To make things easier, I'll split the PR and try find a proper way to solve it.

mprobst:
@mlaval Is this still current now?

mlaval:
With the latest changes, there are still some failures, but not the same ones.
The fixes here are no longer relevant, so this PR can be closed.
I'll at the remaining problems later.

PR #201, opened 2015-06-16
feat(facade): add support for Array.pop()

vicb:
@mprobst if I got it right, that was an easy one


mprobst:
LGTM, please merge.

And yes, these things are supposed to be easy :-)

PR #202, opened 2015-06-17
feat(facade): add support for Array shift and unshift

vicb:


lib/facade_converter.ts
```diff
@@ -93,6 +93,17 @@ export class FacadeConverter extends base.TranspilerBase {
...
       this.emitCall('removeLast');
     },
+    'Array.shift': (c: ts.CallExpression, context: ts.Expression) => {
+      this.visit(context);
+      this.emit('. removeAt ( 0 )');
+    },
+    'Array.unshift': (c: ts.CallExpression, context: ts.Expression) => {
+      this.emit('(');
+      this.visit(context);
+      this.emit('.. insertAll ( 0, [');
```
mprobst: Nit: I wonder if it makes sense to special case the `arguments.length == 1` case? It will give more idiomatic Dart code, and would also save allocating an array and produce shorter code. Might not matter once the JIT went over it, but something to consider, in particular as TypeScript code authors don't really have a way to write the "better" version manually..

mprobst:
See comment, otherwise LGTM, please merge.

vicb:
Thanks for review, comment addressed & landed as 017024a21935c7d435aeedc9ef776b65821f59c7

PR #205, opened 2015-06-18
feat(facade): add support for Map.delete(k)

vicb:
**THIS SHOULD NOT BE MERGED AS IS, THERE IS AN ERROR IN TESTS**

@mprobst I'm not sure to understand the error here, isn't it complaining about x not being typed ?

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -213,6 +213,20 @@ export class FacadeConverter extends base.TranspilerBase {
         this.visit(context);
         this.emitCall('containsKey', c.arguments);
       },
+      'Map.delete': (c: ts.CallExpression, context: ts.Expression) => {
+        // JS Map.delete(k) returns whether k was present in the map,
+        // convert to:
+        // (Map.containsKey(k) && (Map.remove(k) != null || true))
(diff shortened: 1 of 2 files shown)


mprobst:
@vicb the problem is that in our fake externs within `facade_converter_test.ts`, there is no declaration for `Map.delete`, and so the TypeScript type check fails. I should take a look and see why it doesn't give a proper error message.

The fix is to declare the `delete` method at the beginning of the file. Maybe just copy&paste everything from the Angular code base, this has bitten me before, too.

vicb:
Thanks for the tip.

I had an other look and `JS.delete` returns whether the key was actually present, is it reasonable to convert `.delete` to an IIFE ?

mprobst:
Should you be able to just return `myMap.containsKey(k) && (myMap.remove(k) != null || true)`? The latter will always be `true`, but with defined execution order it must always be executed, so together this returns whether `k` was in the map.

vicb:
Nice trick !

vicb:
PR update, PTAL.

mprobst:
LGTM, please merge.

vicb:
Thanks for the review

PR #208, opened 2015-06-18
feat: support façade-handling properties.

### LINKED ISSUE #206: Add support for properties for facade converter
JS has `Map.size` while Dart has `Map.length`.

It seems like properties are not supported, only methods, right ?

/cc @mprobst

mprobst:
Fixes #206.

### DIFF
--- a/lib/expression.ts
+++ b/lib/expression.ts
@@ -68,14 +68,15 @@ class ExpressionTranspiler extends base.TranspilerBase {
 
       case ts.SyntaxKind.PropertyAccessExpression:
         var propAccess = <ts.PropertyAccessExpression>node;
-        this.visit(propAccess.expression);
         if (propAccess.name.text === 'stack' &&
             this.hasAncestor(propAccess, ts.SyntaxKind.CatchClause)) {
           // Handle `e.stack` accesses in catch clauses by mangling to `e_stack`.
           // FIXME: Use type checker/FacadeConverter to make sure this is actually Error.stack.
+          this.visit(propAccess.expression);
           this.emitNoSpace('_stack');
         } else {
-          this.fc.checkPropertyAccess(propAccess);
+          if (this.fc.handlePropertyAccess(propAccess)) break;
+          this.visit(propAccess.expression);
           this.emit('.');
           this.visit(propAccess.name);
         }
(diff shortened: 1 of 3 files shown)


lib/facade_converter.ts
```diff
@@ -3,22 +3,27 @@ import base = require('./base');
...
-          .filter((k) => this.subs[fileName].hasOwnProperty(k))
+    this.extractPropertyNames(this.callHandlers);
+    this.extractPropertyNames(this.propertyHandlers);
+  }
+
+  private extractPropertyNames(m: ts.Map<ts.Map<any>>) {
+    for (var fileName in m) {
+      Object.keys(m[fileName])
+          .filter((k) => m[fileName].hasOwnProperty(k))
           .map((fnName) => fnName.substring(fnName.lastIndexOf('.') + 1))
```
vicb: nit: fnName -> propName ?

lib/facade_converter.ts
```diff
@@ -190,34 +233,13 @@ export class FacadeConverter extends base.TranspilerBase {
...
-
-  private isConstCall(node: ts.CallExpression): boolean {
-    return node && base.ident(node.expression) === 'CONST_EXPR';
-  }
+  private propertyHandlers: ts.Map<ts.Map<PropertyHandler>> = {
+    'angular2/traceur-runtime': {
+      'Map.size': (p: ts.PropertyAccessExpression) => {
+        this.visit(p.expression);
+        this.emit('.');
+        this.emit('length');
```
vicb: :+1:

lib/expression.ts
```diff
@@ -68,12 +68,15 @@ class ExpressionTranspiler extends base.TranspilerBase {
 
       case ts.SyntaxKind.PropertyAccessExpression:
         var propAccess = <ts.PropertyAccessExpression>node;
-        this.visit(propAccess.expression);
         if (propAccess.name.text === 'stack' &&
             this.hasAncestor(propAccess, ts.SyntaxKind.CatchClause)) {
+          // Handle `e.stack` accesses in catch clauses by mangling to `e_stack`.
+          // FIXME: Use type checker/FacadeConverter to make sure this is actually Error.stack.
+          this.visit(propAccess.expression);
           this.emitNoSpace('_stack');
```
vicb: - what about adding a test here ?
- could this be solved by adding a transformer that visits the catch clause ? We should be able to determine if `.stack` is accessed on the exception ?

lib/facade_converter.ts
mprobst: Done.

vicb:
a few comments, otherwise LGTM

lib/expression.ts
mprobst: There is a test for this behaviour (in `test/statement_test.ts`), note that I'm only rewording the comment and slightly refactoring the code.

mprobst:
Thanks for the review!

PR #211, opened 2015-06-21
fix: add multiple arguments to Array.concat

vicb:
@mprobst let's discuss next week on how to add a type test

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -170,9 +170,16 @@ export class FacadeConverter extends base.TranspilerBase {
     'Array.concat': (c: ts.CallExpression, context: ts.Expression) => {
       this.emit('new List . from (');
       this.visit(context);
-      this.emit(') .. addAll (');
-      this.visit(c.arguments[0]);
       this.emit(')');
+      c.arguments.forEach(arg => {
+        var symbol = this.tc.getTypeAtLocation(arg).getSymbol();
+        if (!symbol || this.tc.getFullyQualifiedName(symbol) !== 'Array') {
+          this.reportError(arg, 'Array.concat only supports Array args');
+        }
+        this.emit('.. addAll (');
+        this.visit(arg);
+        this.emit(')');
+      });
     },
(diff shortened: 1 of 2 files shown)


mprobst:
As discussed offline, I think the approach to try here is taking `getTypeAtLocation` for each parameter, then retrieving its `Symbol` via `getSymbol()`, and checking if that is the well known `Array` from `lib.d.ts`.

This is of course all going against the spirit of TypeScript's structural type system a bit, but matches Dart's nominal type system.

vicb:
@mprobst I fail to see how it can help ??

The closest I can find to what we need is https://github.com/Microsoft/TypeScript/blob/644ea82ee39f16102b776f98a962570d9dd83475/src/compiler/checker.ts#L5041 but it doesn't seem to be exposed.

vicb:
@mprobst fixed ! Thank you very much for helping me on this.

lib/facade_converter.ts
```diff
@@ -170,9 +170,16 @@ export class FacadeConverter extends base.TranspilerBase {
     'Array.concat': (c: ts.CallExpression, context: ts.Expression) => {
       this.emit('new List . from (');
       this.visit(context);
-      this.emit(') .. addAll (');
-      this.visit(c.arguments[0]);
       this.emit(')');
+      c.arguments.forEach(arg => {
+        var symbol = this.tc.getTypeAtLocation(arg).getSymbol();
+        if (!symbol || this.tc.getFullyQualifiedName(symbol) !== 'Array') {
```
mprobst: This is not quite enough. You'll also have to make sure that this is the `Array` symbol that's the globally defined `Array`, the one from `lib.d.ts`. So check if the location where `symbol` is declared is in that particular file.

I don't quite know how to explain this better, I feel like I'm not adding much to what I wrote before. If it's ok for you I can take over the PR?

vicb: I was expecting that checking the FQN would be enough. But yes plus take over.

mprobst:
Pushed as a9fba93.

PR #213, opened 2015-06-22
[WIP] fix(facade): Array.forEach only support a single argument

vicb:
[Review on Reviewable](https://reviewable.io/reviews/angular/ts2dart/213)

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -174,6 +174,14 @@ export class FacadeConverter extends base.TranspilerBase {
       this.visit(c.arguments[0]);
       this.emit(')');
     },
+    'Array.forEach': (c: ts.CallExpression, context: ts.Expression) => {
+      const callback = <ts.FunctionExpression>c.arguments[0];
+      if (callback.parameters.length !== 1) {
+        this.reportError(c, 'Array.forEach callback supports a single argument only');
+      }
+      this.visit(context);
+      this.emitCall('forEach', [callback]);
+    },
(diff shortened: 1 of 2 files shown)


vicb:
WIP: I should add support for `A.forEach(fn)` (ie not using an inline fn but referencing a fn)

lib/facade_converter.ts
```diff
@@ -174,6 +174,14 @@ export class FacadeConverter extends base.TranspilerBase {
       this.visit(c.arguments[0]);
       this.emit(')');
     },
+    'Array.forEach': (c: ts.CallExpression, context: ts.Expression) => {
+      const callback = <ts.FunctionExpression>c.arguments[0];
+      if (callback.parameters.length !== 1) {
+        this.reportError(c, 'Array.forEach callback supports a single argument only');
```
mprobst: nit: `takes exactly one argument`?

test/facade_converter_test.ts
```diff
@@ -132,9 +136,18 @@ describe('type based translation', () => {
   });
 
   describe('error detection', () => {
+    describe('Array', () => {
+      it('.forEach should error on multiple arguments', () => {
+        chai.expect(() => translateSource('var x: Array<number> = []; x.forEach((v, i) => v);',
```
mprobst: `expectErroneousWithType` ?

vicb: Ah that was it means :)
I'll update.

PR #215, opened 2015-06-22
[WIP] feat(Facade): add support for Map.forEach

vicb:
[Review on Reviewable](https://reviewable.io/reviews/angular/ts2dart/215)

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -244,6 +244,16 @@ export class FacadeConverter extends base.TranspilerBase {
         this.emitCall('remove', c.arguments);
         this.emit('!= null || true ) )');
       },
+      'Map.forEach': (c: ts.CallExpression, context: ts.Expression) => {
+        const callback = <ts.FunctionExpression>c.arguments[0];
+        let params = callback.parameters;
+        if (params.length !== 2) {
+          this.reportError(c, 'Map.forEach callback supports two arguments only');
+        }
+        [params[0], params[1]] = [params[1], params[0]];
+        this.visit(context);
+        this.emitCall('forEach', [callback]);
+      },
     },
     'angular2/src/di/forward_ref': {
       'forwardRef': (c: ts.CallExpression, context: ts.Expression) => {
--- a/test/facade_converter_test.ts
+++ b/test/facade_converter_test.ts
@@ -11,6 +11,7 @@ var traceurRuntimeDeclarations = `
       set(key: K, value: V): Map<K, V>;
       size: number;
       delete(key: K): boolean;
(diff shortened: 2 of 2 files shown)


lib/facade_converter.ts
```diff
@@ -244,6 +244,16 @@ export class FacadeConverter extends base.TranspilerBase {
         this.emitCall('remove', c.arguments);
         this.emit('!= null || true ) )');
       },
+      'Map.forEach': (c: ts.CallExpression, context: ts.Expression) => {
+        const callback = <ts.FunctionExpression>c.arguments[0];
+        let params = callback.parameters;
+        if (params.length !== 2) {
+          this.reportError(c, 'Map.forEach callback supports two arguments only');
+        }
+        [params[0], params[1]] = [params[1], params[0]];
```
vicb: @mprobst I would prefer creating a new Node instead of mutating an existing one. What do you think ?

vicb:
WIP: when the callback is not an inlined fn (ie references a function) then this should be compiled to 
`m.forEach((k, v) => callback(v, k));

lib/facade_converter.ts
mprobst: You cannot modify the AST, it must be treated as read only - there's internal state in TypeScript that depends on that. So I'm afraid you'll have to manually handle the callback here, and cannot rely on `visit` doing the right thing.

lib/facade_converter.ts
```diff
@@ -244,6 +244,16 @@ export class FacadeConverter extends base.TranspilerBase {
         this.emitCall('remove', c.arguments);
         this.emit('!= null || true ) )');
       },
+      'Map.forEach': (c: ts.CallExpression, context: ts.Expression) => {
+        const callback = <ts.FunctionExpression>c.arguments[0];
+        let params = callback.parameters;
+        if (params.length !== 2) {
+          this.reportError(c, 'Map.forEach callback supports two arguments only');
```
mprobst: nit: `requires two arguments`?

lib/facade_converter.ts
```diff
@@ -244,6 +244,16 @@ export class FacadeConverter extends base.TranspilerBase {
         this.emitCall('remove', c.arguments);
         this.emit('!= null || true ) )');
       },
+      'Map.forEach': (c: ts.CallExpression, context: ts.Expression) => {
+        const callback = <ts.FunctionExpression>c.arguments[0];
```
mprobst: What about the case of passing an already bound expression? `map.forEach(myFn)`? I think it's ok to given an error, but we should have a proper message.

vicb: I plan to update, see the PR comment

PR #216, opened 2015-06-22
fix: facade converter on Windows

mlaval:
A fix so that the facade converter in correctly executed on Windows.

[Review on Reviewable](https://reviewable.io/reviews/angular/ts2dart/216)

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -91,7 +91,7 @@ export class FacadeConverter extends base.TranspilerBase {
 
     var fileName = symbol.valueDeclaration.getSourceFile().fileName;
     fileName = this.getRelativeFileName(fileName);
-    fileName = fileName.replace(/(\.d)?\.ts$/, '');
+    fileName = fileName.replace(/(\.d)?\.ts$/, '').replace(/\\/g, '/');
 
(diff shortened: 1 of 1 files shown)


mlaval:
The build failure seems to be a flake, but I can't restart.
@mprobst PTAL, this is needed to have a proper dev environment for Angular2 on Windows

lib/facade_converter.ts
```diff
@@ -91,7 +91,7 @@ export class FacadeConverter extends base.TranspilerBase {
 
     var fileName = symbol.valueDeclaration.getSourceFile().fileName;
     fileName = this.getRelativeFileName(fileName);
-    fileName = fileName.replace(/(\.d)?\.ts$/, '');
+    fileName = fileName.replace(/(\.d)?\.ts$/, '').replace(/\\/g, '/');
```
mprobst: Shouldn't `getSourceFile().fileName` already only contain `/` from the call to `createSourceFile` in `main.ts`?

Sprinkling the `.replace` all over the place will probably just make the problem worse. We should make sure that all paths that flow through the system use `/`.

mlaval:
Closing in favor of #224

PR #219, opened 2015-06-23
fix(RegExp): properly escape single quotes

vicb:
`replace(string)` replaces the first occurrence only


mprobst:
LGTM, please merge.

vicb:
Thanks for the review

PR #220, opened 2015-06-23
feat(typedefs): support typedefs; unsupport inline typedefs; format

### LINKED ISSUE #9: Support function types (TypeScript), a.k.a. typedefs (Dart)
This:

```
interface SearchFunc {
  (source: string, subString: string): boolean;
}
```

Should transpile to this:

```
typedef bool SearchFunc(String source, String subString);
```

yjbanov:
fixes #9

### DIFF
--- a/lib/declaration.ts
+++ b/lib/declaration.ts
@@ -29,7 +29,14 @@ class DeclarationTranspiler extends base.TranspilerBase {
         break;
       case ts.SyntaxKind.InterfaceDeclaration:
         var ifDecl = <ts.InterfaceDeclaration>node;
-        this.visitClassLike('abstract class', ifDecl);
+        // Function type interface in an interface with a single declaration
+        // of a call signature (http://goo.gl/ROC5jN).
+        if (ifDecl.members.length === 1 && ifDecl.members[0].kind === ts.SyntaxKind.CallSignature) {
+          this.visitCallSignature(ifDecl.name.text, <ts.SignatureDeclaration>ifDecl.members[0],
+                                  ifDecl.typeParameters);
+        } else {
+          this.visitClassLike('abstract class', ifDecl);
+        }
         break;
       case ts.SyntaxKind.HeritageClause:
         var heritageClause = <ts.HeritageClause>node;
@@ -438,6 +445,22 @@ class DeclarationTranspiler extends base.TranspilerBase {
                          'initializers for named parameters must be empty object literals');
       }
     }
+  }
+
+  private visitCallSignature(name: string, signature: ts.SignatureDeclaration,
+                             typeParameters: ts.NodeArray<ts.TypeParameterDeclaration>) {
+    this.emit('typedef');
+    if (signature.type) {
+      this.visit(signature.type);
+    }
+    this.emit(name);
+    if (typeParameters) {
+      this.emit('<');
+      this.visitList(typeParameters);
+      this.emit('>');
+    }
+    this.visitParameters(signature.parameters);
+    this.emit(';');
   }
 }
 
--- a/test/declaration_test.ts
+++ b/test/declaration_test.ts
@@ -133,6 +133,17 @@ describe('interfaces', () => {
   });
 });
 
+describe('single call signature interfaces', () => {
+  it('should support declaration', () => {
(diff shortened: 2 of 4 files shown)


lib/declaration.ts
```diff
@@ -439,6 +446,22 @@ class DeclarationTranspiler extends base.TranspilerBase {
       }
     }
   }
+
+  private visitCallSignature(name: string, signature: ts.SignatureDeclaration,
+                             typeParameters: ts.NodeArray<ts.TypeParameterDeclaration>) {
+    this.emit('typedef');
+    if (signature.type) {
```
mprobst: What would this be here? If I understand this correctly, it'd be the resolved type represented by the interface declaration; I don't think we'd want to visit that here, would we?

lib/type.ts
```diff
@@ -45,9 +45,8 @@ class TypeTranspiler extends base.TranspilerBase {
         this.emit('>');
         break;
       case ts.SyntaxKind.FunctionType:
-        this.emit('dynamic /*');
-        this.emit(node.getText());
-        this.emit('*/');
+        this.reportError(node, 'Inline function type declarations not supported. ' +
```
mprobst: Please don't do this.

test/declaration_test.ts
```diff
@@ -133,6 +133,17 @@ describe('interfaces', () => {
   });
 });
 
+describe('function types', () => {
```
mprobst: The tests usually describe the TypeScript construct that's supported, so this should be `describe('single call signature interfaces', () ...)`.

lib/declaration.ts
yjbanov: This is the return type in the signature. The only things I'm hoisting in from the interface declaration are the generic type parameters and the name. My tests wouldn't pass otherwise, I think.

mprobst:
Thanks for doing this! Please revert the regression for the function types, otherwise this is good to go (with a couple of nits).

lib/declaration.ts
mprobst: Ah I see, makes sense.

lib/type.ts
yjbanov: It does not support anonymous function type for local non-parameter variables (more on that below). However, every anonymous function type declaration can also be expressed via a private _named_ function type alias, so this is not a regression but merely a redirection to use another way of doing the same thing, one that ts2dart is able to translate.

Alternatives are:

- (status quo) allow anonymous function types but erase them during transpilation: this is bad because users will feel that they can use this while they shouldn't, otherwise Dart code quality will hurt
- generate a named function type: however this will require that we generate a name for it, which is unlikely to be meaningful. Again, Dart code quality will hurt.

However, Dart does support anonymous type aliases for parameters. I'll see if I can squeeze in another pull request for that. This what they look like:

```dart
func(void fn(int));
```

test/declaration_test.ts
yjbanov: Ok, but in the docs on their web-site they refer to them as "function type"

mprobst: The interface does describe a function type and is isomorphic with `(a: string) => number`, but the syntactical form is an interface with a single call signature - which is what we should be using as terminology here, as we're doing syntax directed translation and mostly ignore the actual types.

lib/type.ts
mprobst: This affects our user base and developers, so I think we should discuss this in a larger group, and not just declare facts by putting code in. Let's discuss what we want here in the standup.

I think if we enable the parameter function types, we're covering most of our use cases.

Also, is there any intention in Dart to support general function types?

yjbanov: I'll start a separate PR for this

test/declaration_test.ts
yjbanov: Done

mprobst:
Pushed as 017024a21935c7d435aeedc9ef776b65821f59c7.

yjbanov:
Perhaps pushed as https://github.com/angular/ts2dart/commit/ee6b27dbbe9d2caf4705089f6c5d67d8a7bcfa62? :)

mprobst:
Indeed. I'm 100% sure I copied the SHA from the page, maybe github has some odd caching issue there?

PR #222, opened 2015-06-23
feat(types): disallow inline function types

yjbanov:
This PR is a spin-off of https://github.com/angular/ts2dart/pull/220 (more discussion: https://github.com/angular/ts2dart/pull/220#discussion_r33064068)

Inline function types are not supported in Dart for local variables other than function parameters and allowing them in ts2dart gives developers a false impression that they can use it, while in fact dart2js simply erases them producing bad Dart code. Given the goal of the project is to produce idiomatic Dart, it is reasonable to push developers to use only those language constructs that ts2dart is able to translate cleanly, which in this case is named function types.

### DIFF
--- a/lib/literal.ts
+++ b/lib/literal.ts
@@ -33,7 +33,7 @@ class LiteralTranspiler extends base.TranspilerBase {
         if (tmpl.templateSpans) this.visitEach(tmpl.templateSpans);
         break;
       case ts.SyntaxKind.TemplateHead:
-        this.emit(`'''${this.escapeTextForTemplateString(node)}`); //highlighting bug:'
+        this.emit(`'''${this.escapeTextForTemplateString(node)}`);  // highlighting bug:'
         break;
       case ts.SyntaxKind.TemplateTail:
         this.emitNoSpace(this.escapeTextForTemplateString(node));
--- a/lib/type.ts
+++ b/lib/type.ts
@@ -45,9 +45,8 @@ class TypeTranspiler extends base.TranspilerBase {
         this.emit('>');
         break;
       case ts.SyntaxKind.FunctionType:
-        this.emit('dynamic /*');
-        this.emit(node.getText());
-        this.emit('*/');
+        this.reportError(node, 'Inline function type declarations not supported. ' +
+                                   'Use function types instead (http://goo.gl/ROC5jN).');
         break;
       case ts.SyntaxKind.QualifiedName:
         var first = <ts.QualifiedName>node;
--- a/test/literal_test.ts
+++ b/test/literal_test.ts
@@ -4,7 +4,7 @@ import {expectTranslate, expectErroneousCode} from './test_support';
 describe('literals', () => {
   it('translates string literals', () => {
     expectTranslate(`'hello\\' "world'`).to.equal(` "hello' \\"world" ;`);
-      expectTranslate(`"hello\\" 'world"`).to.equal(` "hello\\" 'world" ;`);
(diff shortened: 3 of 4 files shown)


mprobst:
I discussed this with /cc @mhevery, too, who asked how much of a problem this is.

Looking through our API, I cannot find instances of actually declaring variables or structs with function types that show up in our API. I do find quite a bunch of places where we declare a parameter as an inline function type.

I think our efforts would be better spent here enabling our users to use function types in parameters than outlawing something that's not actually a problem.

So what about sending a PR that adds support for function types in function parameters? It should be easy enough to special case by checking what the parent of a function type node is.

yjbanov:
I'll write one for parameters when I get a chance. I would still like to outlaw inline typedefs for local variables. There may be a reasonable strategy for them, but erasing them is just not it.

mprobst:
@yjbanov I understand that's your opinion (fn typedefs not being acceptable, or more generally, not using TypeScript features that are not supported by Dart). But could you explain your data or reasoning?

I cannot see a way how inline variables with function types (non-API) hurt Dart users in any way. Allowing them will make Angular developers more productive (more idiomatic TS code, fewer errors late in the development cycle from ts2dart, and it's a nice feature in general). So overall this is a win for Dart users, too.

yjbanov:
I do not understand your logic. You seem to simultaneously claim that this feature is valuable and at the same time argue that it's ok to erase it from Dart code. Remember, it is a requirement that the Dart code we output is idiomatic. There's tons of stuff in TS you cannot use because it won't translate well to Dart. It's one of the big pain points in our development cycle. One has to make it work in one language, and then the other, often in multiple iterations. Removing a feature for which _you have an alternative_ is not a problem. On the contrary, it allows you to catch errors early.

Specifically on function types, there is a way to express what you want in an idiomatic way in both languages. Therefore it is beneficial for us to not allow code that would result in non-idiomatic code. 

## Example

This:

```typescript
var x: (a: string) => string;
if (makeAllCaps) {
  x = toUpperCase;
} else {
  x = toLowerCase;
}
format(x);
```

can be easily expressed like this:

```typescript
interface Formatter { (a: string) => string; }

var x: Formatter;
if (makeAllCaps) {
  x = toUpperCase;
} else {
  x = toLowerCase;
}
format(x);
```

The latter gets you proper type checking in both Dart and in TS. The former doesn't. I'm sure there are corner cases in the type system that may result in one language accepting a typedef and the other one rejecting it. We should surface this as early as possible.

Feel free to close the issue. It's not terribly important. I'm worried about death by a 1000 cuts we might find if we're not careful with our decisions over time.

mprobst:
I am closing the issue, but your comment does deserve a reply.

> I do not understand your logic. You seem to simultaneously claim that this feature is valuable and at the same time argue that it's ok to erase it from Dart code.

Dart users use Angular's API; nobody develops Angular by using Dart as their primary language. Thus it is ok to use valuable features of TypeScript as long as they don't (strongly) impact the API.

> Remember, it is a requirement that the Dart code we output is idiomatic.

You put this as an absolute, but that's a fallacy - if we wanted absolutely idiomatic Dart over all else, we'd be writing in Dart. We don't, so the level of "idiomacity" we're aiming for is a relative, not an absolute. Using a function type in a local variable is not a problem - it doesn't affect Dart users enough to be a problem.

> Removing a feature for which you have an alternative is not a problem. On the contrary, it allows you to catch errors early.

Like, only programming using `NAND`? :-) 

If you do find something is a problem in our public API for Dart users, please open an issue for that.

PR #224, opened 2015-06-24
feat: better Windows path handling

mprobst:
/cc @mlaval

[Review on Reviewable](https://reviewable.io/reviews/angular/ts2dart/224)

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -80,10 +80,11 @@ export class Transpiler {
    * @param destination Location to write files to. Creates files next to their sources if absent.
    */
   transpile(fileNames: string[], destination?: string): void {
-    var host = this.createCompilerHost(fileNames);
     if (this.options.basePath) {
       this.options.basePath = this.normalizeSlashes(this.options.basePath);
     }
+    fileNames = fileNames.map((f) => this.normalizeSlashes(f));
+    var host = this.createCompilerHost();
     if (this.options.basePath && destination === undefined) {
       throw new Error('Must have a destination path when a basePath is specified ' +
                       this.options.basePath);
@@ -95,12 +96,12 @@ export class Transpiler {
     }
 
     // Only write files that were explicitly passed in.
-    var fileMap: {[s: string]: boolean} = {};
-    fileNames.forEach((f) => fileMap[this.normalizeSlashes(f)] = true);
+    var fileSet: {[s: string]: boolean} = {};
+    fileNames.forEach((f) => fileSet[f] = true);
 
     this.errors = [];
     program.getSourceFiles()
-        .filter((sourceFile) => fileMap[sourceFile.fileName])
+        .filter((sourceFile) => fileSet[sourceFile.fileName])
         // Do not generate output for .d.ts files.
         .filter((sourceFile: ts.SourceFile) => !sourceFile.fileName.match(/\.d\.ts$/))
         .forEach((f: ts.SourceFile) => {
@@ -133,8 +134,9 @@ export class Transpiler {
     return opts;
   }
 
-  private createCompilerHost(files: string[]): ts.CompilerHost {
+  private createCompilerHost(): ts.CompilerHost {
     var defaultLibFileName = ts.getDefaultLibFileName(COMPILER_OPTIONS);
+    defaultLibFileName = this.normalizeSlashes(defaultLibFileName);
     return {
       getSourceFile: (sourceName, languageVersion) => {
         var path = sourceName;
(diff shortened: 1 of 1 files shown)


mprobst:
@mlaval could you patch this in on a Windows machine and see if it works for you?

If this gives errors about not being able to load files, could you uncomment [main.ts#L144] and see if that fixes it?

mlaval:
It doesn't solve the issue fixed by #216 :( But it doesn't seem to break anything new.
It seems that TS outputs a path with backward slashes anyway.

On a side note, `if (path.resolve('/', filePath) !== filePath)` doesn't replace path.isAbsolute on Windows.

mprobst:
Try again?

Also, could you be more specific on how the `path.resolve` thing doesn't work as intended? Do you mean `path.resolve('/', 'C:\\x') !== 'C:\\x'` even though that should be absolute? Might have to move to node v0.12 then...

mlaval:
Same results with this version.

For the absolute check:
- `path.resolve('/', 'C:\\x') === 'C:\\x'`
- `path.resolve('/', 'C:/x') === 'C:\\x'`  
And paths were already converted to forward slash at this point, hence the previous workaround:
`if (filePath.indexOf('/') !== 0 && filePath.indexOf(':/') !== 1)`
`path.isAbsolute` would solve that if we hade Node 0.12

mprobst:
That's the expected result, I want `path.resolve(..., C:\)` and `path.resolve(..., C:\)` to just result in `C:\`?

mprobst:
OK, pushed some code that should fix that, too.

lib/main.ts
```diff
@@ -135,14 +135,18 @@ export class Transpiler {
...
       getSourceFile: (sourceName, languageVersion) => {
         var path = sourceName;
         if (sourceName === defaultLibFileName) {
           path = ts.getDefaultLibFilePath(COMPILER_OPTIONS);
         }
+        // if (process.platform === 'win32') path = sourceName.replace(/\//g, '\\');
         if (!fs.existsSync(path)) return undefined;
         var contents = fs.readFileSync(path, 'UTF-8');
+        // Make sure the file path is relative and all file names use forward slashes.
+        var sourceName = this.getRelativeFileName(sourceName);
```
mlaval: Passing the relative path here generates some failures locally when running the Dart unit tests of Angular2. It seems that the file cache of TS is going wrong (something like picking the original source instead of a transformed file from a Broccoli tree):
```
[00:33:07] Error: [TSToDartTranspiler]: Debug Failure. False expression: end mus
t be the bounds of the file. 531 > 373
    at Object.assert (c:\Dev\angular_mlaval\node_modules\ts2dart\node_modules\typescript\bin\typescript.js:1379:23)
    at Object.createFileDiagnostic (c:\Dev\angular_mlaval\node_modules\ts2dart\node_modules\typescript\bin\typescript.js:999:15)
    at getSourceFileFromCache (c:\Dev\angular_mlaval\node_modules\ts2dart\node_m
odules\typescript\bin\typescript.js:29901:44)
    at findSourceFile (c:\Dev\angular_mlaval\node_modules\ts2dart\node_modules\typescript\bin\typescript.js:29861:24)
    at processSourceFile (c:\Dev\angular_mlaval\node_modules\ts2dart\node_module
s\typescript\bin\typescript.js:29831:27)
    at c:\Dev\angular_mlaval\node_modules\ts2dart\node_modules\typescript\bin\typescript.js:29910:17
    at Object.forEach (c:\Dev\angular_mlaval\node_modules\ts2dart\node_modules\typescript\bin\typescript.js:718:30)
    at processReferencedFiles (c:\Dev\angular_mlaval\node_modules\ts2dart\node_modules\typescript\bin\typescript.js:29908:16)
    at findSourceFile (c:\Dev\angular_mlaval\node_modules\ts2dart\node_modules\t
ypescript\bin\typescript.js:29884:25)
    at processSourceFile (c:\Dev\angular_mlaval\node_modules\ts2dart\node_modules\typescript\bin\typescript.js:29831:27)
```

Going back to `var sourceName = this.normalizeSlashes(sourceName);` solves this issue and allows to run the tests successfully :)

mprobst:
@mlaval I'm afraid if we're going back to `normalizeSlashes`, we'll end up with non-relative paths again. That it works might just be a random fluke because a backslashed path is the same length as a forward-slashed.

I'm not sure how to make progress here, it seems too cumbersome to remote run this with on Windows and me on different platforms. While I'm super happy to get Windows support if possible, sprinkling calls to `normalizePath` all over the code base is a no-no.

Maybe you could see if you can find an elegant solution, confined only to `main.ts` and the source host? Otherwise I'd consider rather dropping Windows support (for developers), I'm not sure if it's worth the effort.

mlaval:
The error I encountered with the relative path is not related to Windows, I have the very same one on a Linux VM. So it seems to be a separate topic.

To make things work on Windows, only the 3 `normalizeSlashes` are needed in `main.ts`: for the `defaultLibFileName`, in `getOutputPath` and in `getRelativeFileName`.

Let's discuss that next week.

mprobst:
Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/angular/ts2dart/224)

mprobst:
@mlaval PTAL, and please test.

PR #225, opened 2015-06-26
refactor(tests): use expectErroneousWithType()

vicb:
[Review on Reviewable](https://reviewable.io/reviews/angular/ts2dart/225)

### DIFF
--- a/test/facade_converter_test.ts
+++ b/test/facade_converter_test.ts
@@ -136,7 +136,7 @@ describe('type based translation', () => {
   describe('error detection', () => {
     describe('Array', () => {
       it('.concat() should report an error if any arg is not an Array', () => {
-        chai.expect(() => translateSource('var x: Array<number> = []; x.concat(1);', COMPILE_OPTS))
+        expectErroneousWithType('var x: Array<number> = []; x.concat(1);')
             .to.throw('Array.concat only takes Array arguments');
       });
     });


vicb:
Not sure what happens with clang-format. It want me to reformat literal.ts to align "//" comments. Is this something we really want ?

Also notes than Travis fails with 403 even after multiple trials (I've created a separate issue for that)

mprobst:
LGTM if Travis is green (or confirmed only flaking).

mprobst:
Regarding aligning line comments (`//`), I don't really care much, but I guess it can improve "scanability" of many end of line comments in a row. No idea why `literal.ts` isn't properly formatted though, maybe it's a versioning issue?

vicb:
I see no reason to align eol comments (but I don't have anything strong against doing so).

The real question here is why clang-format is ok on Travis while it is not on my machine. This will cause troubles (when ok locally but not on Travis) and must be fixed

vicb:
```
[15:00:23] WARNING: Files are not properly formatted. Please run
[15:00:23]   ./node_modules/gulp-clang-format/node_modules/clang-format/index.js -i -style="file" /home/victor/dart/ts2dart/lib/literal.ts
[15:00:24]   (using clang-format version 1.0.25)
[15:00:24] files not formatted
```

I'll merge this PR and create a separate issue to track this

vicb:
landed as 25a38753066afb071ec0490348340d2bc016fe71

PR #231, opened 2015-06-30
feat(facade): add XMLHttpRequest -> HttpRequest to typename

jeffbcross:
Closes #230

[Review on Reviewable](https://reviewable.io/reviews/angular/ts2dart/231)

Issue #230, opened 2015-06-29
add support for XMLHttpRequest to base.ts (use HttpRequest in Dart)


jeffbcross:

mprobst:
@jeffbcross what do you mean exactly by support exactly? `XMLHttpRequest` AFAIK has an API that's very different from `HttpRequest` in Dart. As `XMLHttpRequest`s API is famously wonky, it'd IMHO be better to just introduce a platform specific abstraction.

That is, have a `http_request.ts` for JavaScript that wraps `XMLHttpRequest` in a sane API, and (in the same place) `http_request.dart` that does the same for Dart. Both should export the same symbols/API, and then for Dart the `.dart` gets substituted transparently.

jeffbcross:
Sorry, I opened this during a code review and should've added more detail.

The APIs are actually very similar, and I have an abstraction already ([ts](https://github.com/angular/angular/blob/master/modules/angular2/src/http/backends/browser_xhr.ts), [dart](https://github.com/angular/angular/blob/master/modules/angular2/src/http/backends/browser_xhr.dart)), but I am forced to use ["any" type](https://github.com/angular/angular/blob/5b5ffe75d0d1a3ac6204114203d8d566d49e9c70/modules/angular2/src/http/backends/xhr_backend.ts#L27) on the fields to which instances are assigned. 

I

mprobst:
I see, that's a valid use case (and a bit surprising on the Dart side, why would you emulate `onReadyStateChange`?).

Send a PR to add the type renaming?

Also note that the `declare var window;` in the code you linked to above is superfluous, AFAICT.

jeffbcross:
Sounds good, will open a PR. And will remove the window declaration.

mprobst:
This has been done.

PR #232, opened 2015-07-01
feat(typing): namespace the translated type names.

alexeagle:
This allows us to add more type translations for
DOM API without worry about colliding with other
unrelated types.

This required a bit of refactoring since the
TypeChecker is not generally available to the
transpilers.

[Review on Reviewable](https://reviewable.io/reviews/angular/ts2dart/232)

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -11,19 +11,22 @@ const FACADE_DEBUG = false;
 export class FacadeConverter extends base.TranspilerBase {
   private tc: ts.TypeChecker;
   private candidateProperties: {[propertyName: string]: boolean} = {};
+  private candidateTypes: {[typeName: string]: boolean} = {};
 
   constructor(transpiler: ts2dart.Transpiler) {
     super(transpiler);
     this.extractPropertyNames(this.callHandlers);
     this.extractPropertyNames(this.propertyHandlers);
+    this.extractPropertyNames(this.TS_TO_DART_TYPENAMES, this.candidateTypes);
   }
 
-  private extractPropertyNames(m: ts.Map<ts.Map<any>>) {
+  private extractPropertyNames(m: ts.Map<ts.Map<any>>,
+                               candidates: {[k: string]: boolean} = this.candidateProperties) {
     for (var fileName in m) {
       Object.keys(m[fileName])
           .filter((k) => m[fileName].hasOwnProperty(k))
           .map((propName) => propName.substring(propName.lastIndexOf('.') + 1))
-          .forEach((propName) => this.candidateProperties[propName] = true);
+          .forEach((propName) => candidates[propName] = true);
     }
   }
 
@@ -85,6 +88,30 @@ export class FacadeConverter extends base.TranspilerBase {
     return handler && !handler(pa);
   }
 
+  visitTypeName(typeName: ts.EntityName) {
+    if (typeName.kind !== ts.SyntaxKind.Identifier) {
+      this.visit(typeName);
+      return;
+    }
+    var ident = base.ident(typeName);
+    if (this.candidateTypes.hasOwnProperty(ident) && this.tc) {
+      var symbol = this.tc.getSymbolAtLocation(typeName);
+      if (!symbol) {
+        this.reportMissingType(typeName, ident);
+        return;
+      }
+      let fileAndName = this.getFileAndName(symbol);
+      if (fileAndName) {
+        var fileSubs = this.TS_TO_DART_TYPENAMES[fileAndName.fileName];
+        if (fileSubs && fileSubs.hasOwnProperty(fileAndName.qname)) {
+          this.emit(fileSubs[fileAndName.qname]);
+          return;
+        }
+      }
+    }
+    this.emit(ident);
+  }
+
   private getHandler<T>(symbol: ts.Symbol, m: ts.Map<ts.Map<T>>): T {
     var {fileName, qname} = this.getFileAndName(symbol);
     var fileSubs = m[fileName];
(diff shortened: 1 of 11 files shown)


lib/facade_converter.ts
```diff
@@ -146,6 +181,33 @@ export class FacadeConverter extends base.TranspilerBase {
...
+    // https://github.com/angular/angular/issues/2770
+    'Node': 'dynamic',
+    'Text': 'dynamic',
+    'Element': 'dynamic',
+    'HTMLElement': 'dynamic',
+    'HTMLStyleElement': 'dynamic',
+    'HTMLInputElement': 'dynamic',
+    'HTMLDocument': 'dynamic',
+    'History': 'dynamic',
+    'Location': 'dynamic',
```
jeffbcross: Can you tack on 'XMLHttpRequest': 'HttpRequest' here so I can close PR #231 that would conflict with this change (as well as issue #230)?

lib/facade_converter.ts
```diff
@@ -85,6 +85,31 @@ export class FacadeConverter extends base.TranspilerBase {
     return handler && !handler(pa);
   }
 
+  visitTypeName(typeName: ts.EntityName) {
+    if (typeName.kind !== ts.SyntaxKind.Identifier) {
+      this.visit(typeName);
```
mprobst: Add a comment why this can occur?

lib/facade_converter.ts
```diff
@@ -85,6 +85,31 @@ export class FacadeConverter extends base.TranspilerBase {
...
+    }
+    var ident = base.ident(typeName);
+    if (this.tc) {
+      var symbol = this.tc.getSymbolAtLocation(typeName);
+      if (!symbol) {
+        this.reportMissingType(typeName, ident);
+        return;
+      }
+
+      if (this.getFileAndName(symbol) !== null) {
```
mprobst: extract the value to a var, so you don't have to getFileAndName twice? I expect it to be not entirely cheap.

lib/facade_converter.ts
```diff
@@ -85,6 +85,31 @@ export class FacadeConverter extends base.TranspilerBase {
...
 
+  visitTypeName(typeName: ts.EntityName) {
+    if (typeName.kind !== ts.SyntaxKind.Identifier) {
+      this.visit(typeName);
+      return;
+    }
+    var ident = base.ident(typeName);
+    if (this.tc) {
+      var symbol = this.tc.getSymbolAtLocation(typeName);
+      if (!symbol) {
```
mprobst: You could be a bit less draconic by checking first if the type name is ever substituted at all, and only report an error if a collision could happen - same as we do for property names.

lib/facade_converter.ts
```diff
@@ -93,21 +118,31 @@ export class FacadeConverter extends base.TranspilerBase {
   }
 
   private getFileAndName(symbol: ts.Symbol): {fileName: string, qname: string} {
-    while (symbol.flags & ts.SymbolFlags.Alias) symbol = this.tc.getAliasedSymbol(symbol);
-    if (!symbol.valueDeclaration) return null;
+    while (symbol.flags & ts.SymbolFlags.Alias) {
```
mprobst: no single line while loops? :-(

lib/facade_converter.ts
```diff
@@ -146,6 +181,33 @@ export class FacadeConverter extends base.TranspilerBase {
...
+    'Element': 'dynamic',
+    'HTMLElement': 'dynamic',
+    'HTMLStyleElement': 'dynamic',
+    'HTMLInputElement': 'dynamic',
+    'HTMLDocument': 'dynamic',
+    'History': 'dynamic',
+    'Location': 'dynamic',
+  };
+
+  private TS_TO_DART_TYPENAMES: ts.Map<ts.Map<string>> = {
```
mprobst: Nice!

test/facade_converter_test.ts
```diff
@@ -25,9 +25,11 @@ function getSources(str: string): {[k: string]: string} {
     'angular2/traceur-runtime.d.ts': traceurRuntimeDeclarations,
     'angular2/src/di/forward_ref.d.ts': `
         export declare function forwardRef<T>(x: T): T;`,
+    'angular2/src/facade/async.d.ts': `
```
mprobst: Maybe it'd be nice to do a negative test for one of these, e.g. have a different file that defines Observable and see that it doesn't get renamed?

lib/facade_converter.ts
```diff
@@ -93,21 +118,31 @@ export class FacadeConverter extends base.TranspilerBase {
...
 
   private getFileAndName(symbol: ts.Symbol): {fileName: string, qname: string} {
-    while (symbol.flags & ts.SymbolFlags.Alias) symbol = this.tc.getAliasedSymbol(symbol);
-    if (!symbol.valueDeclaration) return null;
+    while (symbol.flags & ts.SymbolFlags.Alias) {
+      symbol = this.tc.getAliasedSymbol(symbol);
+    }
+    let decl = symbol.valueDeclaration;
+    if (!decl) {
+      // Martin: I started getting a null return value from here, which I suppose didn't happen with
```
mprobst: I got the `valueDeclaration` from a code snippet from @vladima, I think. My understanding is that `valueDeclaration` is the first place where the `Symbol` actually gets defined as a value. If you have a pure declaration, i.e. `declare interface Foo {}`, you just don't have an actual value. Maybe that's the issue? Only a theory though.

lib/facade_converter.ts
alexeagle: done

lib/facade_converter.ts
alexeagle: I don't actually have any idea, this is carried over from the prior implementation

lib/facade_converter.ts
alexeagle: That's very plausible. I'll just add the comment explaining it that way.

lib/facade_converter.ts
alexeagle: done, I was scared by the destructuring

lib/facade_converter.ts
alexeagle: oh, I missed how you did that. nice

lib/facade_converter.ts
alexeagle: oh, leftover from a debugging state where I added a statement in there, will revert

test/facade_converter_test.ts
alexeagle: yes, indeed

lib/facade_converter.ts
alexeagle: added a test case as well

alexeagle:
Merged as 380aae4 - let me know if you have any other comments...

PR #240, opened 2015-07-10
fix(ts2dart): emit members for `protected` property parameters

caitp:
I know it's not very meaningful in dart, but it's useful for TypeScript.

When mixed with dart's module-private semantics, it could work okay for
both.


jeffbcross:
@alexeagle mind taking a look?

alexeagle:
LGTM

jeffbcross:
I'll merge, but will wait until Monday to bump the release, since there's apparently a special process.

alexeagle:
Hmm, not good that we haven't published steps for anyone who needs to do a
release.
But I'm not sure what the special process refers to. It's just npm publish
and then update the shrinkwrap.

PR #241, opened 2015-07-10
fix(ts options): turn on experimentalDecorators flag by default.

rkirov:
Typescript 1.5.3 release candidate throw errors on decorator usage
without this flag, which makes angular2 transpilation really noisy.

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -45,6 +45,7 @@ export interface TranspilerOptions {
 
 export const COMPILER_OPTIONS: ts.CompilerOptions = {
   allowNonTsExtensions: true,
+  experimentalDecorators: true,
   module: ts.ModuleKind.CommonJS,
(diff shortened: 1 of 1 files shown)


lib/main.ts
```diff
@@ -45,6 +45,7 @@ export interface TranspilerOptions {
 
 export const COMPILER_OPTIONS: ts.CompilerOptions = {
   allowNonTsExtensions: true,
+  experimentalDecorators: true,
```
mprobst: Generally LGTM, but shouldn't this be a change in Angular that gets sent here via the options passed in the ctor?

mprobst: Ah wait, we don't actually support that :-) Maybe we should, file an issue?

rkirov: as discussed, keeping this as it is, since don't have a single user who would want experimentalDecorators: false

PR #246, opened 2015-07-17
fix(ts2dart): allow generic type parameters on method calls

alexeagle:


lib/call.ts
```diff
@@ -46,10 +46,9 @@ class CallTranspiler extends base.TranspilerBase {
 
   private visitCall(c: ts.CallExpression) {
     this.visit(c.expression);
-    if (c.typeArguments && c.kind !== ts.SyntaxKind.NewExpression) {
-      this.reportError(c, 'Type arguments are only supported on new calls.');
+    if (c.typeArguments && c.kind === ts.SyntaxKind.NewExpression) {
```
mprobst: Add a comment explaining why we allow them?

alexeagle: done

alexeagle:
merged as cb0d919

PR #247, opened 2015-07-17
fix(HttpRequest): Fix wrong XmlHttpRequestMapping

alfonso-presa:
* fixes https://github.com/angular/ts2dart/issues/245

I've tried to solve it in the best way I could. I've been reviewing the code and it doesn't seem there's a way to globally generate imports at the beginning of the file so I added a `buildImports` method to facade_converter that seemed like the more suitable place to handle type>import mapping.

Thanks!

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -88,6 +88,36 @@ export class FacadeConverter extends base.TranspilerBase {
     return handler && !handler(pa);
   }
 
+  buildImports(sourceFile: ts.SourceFile) {
+    var imports: {[type: string]: boolean} = {};
+
+    function getNodes(sourceFile: ts.SourceFile): ts.Node[] {
+      var nodes: ts.Node[] = [];
+      function allNodes(n: ts.Node){ts.forEachChild(n, n => {
+        nodes.push(n);
+        allNodes(n);
+        return false;
+      })};
+      allNodes(sourceFile);
+      return nodes;
+    }
+
+    var types = getNodes(sourceFile)
+                    .filter(n => n.kind === ts.SyntaxKind.TypeReference)
+                    .map(n => base.ident((<ts.TypeReferenceNode>n).typeName));
+
+
+    for (var iType in types) {
+      var type = types[iType];
+      if (this.TS_TO_DART_TYPE_IMPORTS[type]) {
+        imports[this.TS_TO_DART_TYPE_IMPORTS[type]] = true;
+      }
+    }
+    for (var imp in imports) {
+      this.emit('import "' + imp + '";');
+    }
+  }
+
   visitTypeName(typeName: ts.EntityName) {
     if (typeName.kind !== ts.SyntaxKind.Identifier) {
       this.visit(typeName);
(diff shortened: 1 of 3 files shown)


lib/facade_converter.ts
```diff
@@ -88,6 +88,36 @@ export class FacadeConverter extends base.TranspilerBase {
...
+  buildImports(sourceFile: ts.SourceFile) {
+    var imports: {[type: string]: boolean} = {};
+
+    function getNodes(sourceFile: ts.SourceFile): ts.Node[] {
+      var nodes: ts.Node[] = [];
+      function allNodes(n: ts.Node){ts.forEachChild(n, n => {
+        nodes.push(n);
+        allNodes(n);
+        return false;
+      })};
```
mprobst: I think you're missing a `;` behind the `)`, that's why you get the awkward formatting here.

lib/facade_converter.ts
```diff
@@ -88,6 +88,36 @@ export class FacadeConverter extends base.TranspilerBase {
     return handler && !handler(pa);
   }
 
+  buildImports(sourceFile: ts.SourceFile) {
+    var imports: {[type: string]: boolean} = {};
+
+    function getNodes(sourceFile: ts.SourceFile): ts.Node[] {
+      var nodes: ts.Node[] = [];
```
mprobst: This will be rather expensive for a large file, would it be nicer to pass in a predicate directly? Or maybe just have a more specific `findTypeNames` instead of this very generic function?

lib/facade_converter.ts
```diff
@@ -88,6 +88,36 @@ export class FacadeConverter extends base.TranspilerBase {
     return handler && !handler(pa);
   }
 
+  buildImports(sourceFile: ts.SourceFile) {
+    var imports: {[type: string]: boolean} = {};
```
mprobst: maybe call it `extraImports`?

lib/facade_converter.ts
```diff
@@ -88,6 +88,36 @@ export class FacadeConverter extends base.TranspilerBase {
...
+      allNodes(sourceFile);
+      return nodes;
+    }
+
+    var types = getNodes(sourceFile)
+                    .filter(n => n.kind === ts.SyntaxKind.TypeReference)
+                    .map(n => base.ident((<ts.TypeReferenceNode>n).typeName));
+
+
+    for (var iType in types) {
```
mprobst: I'm a big fan of generalizing code, but I think in this case the code would be simpler and much more readable if we'd just plain special case the `XMLHttpRequest => dart:html` case. What about calling the method `maybeEmitDartHtmlImport` and really just doing that?

In case we find we need this more often, we can still go generic later on. But I'd like to see the second and maybe third case before.

mprobst:
Sorry for not responding on the issue soon enough.

Thanks for the PR! I have a couple of comments (see inline). Feel free to push back if something doesn't make sense.

alfonso-presa:
I totally ser your point about making it more specific in sake of efficiency. I'll refactor it and report back.

Thanks!

alfonso-presa:
Let me know it this new commit matches what you meant in your comments.

It's now bounded to `dart:html`, but instead of hard-coding an string comparison with `XMLHttpRequest` I've left a map of objects (even if there's only one entry right now) because it looks like there are plenty of classes in `dart:html` that may be added later.... and a Map lookup is not that bad in terms of efficiency :-).

mprobst:
Thanks for the contribution! I added a slight simplification to use getChildren to avoid the stop exception.

Otherwise submitted as 6d5ed11.

alfonso-presa:
That's by far way more elegant than what I did. I'm new to TS so sorry for that.

I think line 95 (my little exception "trick") is no longer needed thanks to your refactoring.

Thanks to you!

mprobst:
Good point, I missed that. Fixed in e5b325f.

PR #248, opened 2015-07-18
fix(Array): slice and splice wrong ListWrapper syntax

alfonso-presa:
White trying to use splice I got a syntax error:

```
Expected an identifier
erEach ( ( ) { . ListWrapper.splice ( existingXHRs , 0 , existingXHRs .…
```

I worked it around by writing it other way, but I think this should have fixed it.

If I'm wrong just close this PR... I'll get the message :-).

Thanks!

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -171,8 +171,12 @@ export class FacadeConverter extends base.TranspilerBase {
     return node && base.ident(node.expression) === 'CONST_EXPR';
   }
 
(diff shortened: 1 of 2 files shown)


mprobst:
Looks good, but could you add a test so we don't regress again? Thanks!

alfonso-presa:
Of course :-).

mprobst:
Thanks! Merged as a8bd469.

PR #252, opened 2015-07-28
feat(dartdoc): convert link syntax to dartdoc

alexeagle:
for https://github.com/angular/angular/issues/2072

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -245,7 +245,7 @@ export class Transpiler {
         this.lastCommentIdx = c.pos;
         var text = this.currentFile.text.substring(c.pos, c.end);
         this.emitNoSpace('\n');
-        this.emit(text);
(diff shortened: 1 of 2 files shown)


mhevery:
LGTM

alexeagle:
merged in dd4e994

lib/main.ts
```diff
@@ -259,6 +259,10 @@ export class Transpiler {
   }
 
   private normalizeSlashes(path: string) { return path.replace(/\\/g, '/'); }
+  
+  private translateComment(comment: string): string {
+    return comment.replace(/\{@link ([^\}]+)\}/g, '[$1]');
```
mprobst: Should the inner group `[^\}]+` be non-greedy here? Why does this work for the two-link test case below?

PR #254, opened 2015-08-01
feat: use container based Travis builds.

mprobst:
This appears to reduce build start times substantially (seconds, not minutes!).


alexeagle:
LGTM

PR #255, opened 2015-08-04
feat: adjust to new promise symbol location.

mprobst:
Angular now re-exports Promise from the es6-promise.d.ts definition file. This
means ts2dart has to adjust the location its special casing.


alexeagle:
Is there any benefit to supporting Promise in both locations, during a transition? Or is this already broken and just needs fixed?

LGTM

mprobst:
I'll just bundle the upgrade to this new ts2dart version with the change to re-export promises.

Issue #227, opened 2015-06-26
Facade: create utils methods to get info about a fn


vicb:
It would be great to have some support function get extract some info about a fn:
- return types,
- arguments (nb, types).

Use case to implement `Map.forEach` we need to make sure that the cb take two args. If we pass an inline function (or an arrow function), we need to swap the parameters (v, k -> k, v). If an identifier is passed, we still need to check the underlying function takes 2 args and we need to add an indericction to swap param order `m.forEach(fn) -> m.map((k, v) { fn(v, k); });

What we need to support:
- identifier,
- function expression,
- arrow expression.

mprobst:
Thinking about this again, I think it might be easier to just insert the intermediate function. Detecting if the passed expression is a function, and if so, if it takes two args, and if so, if we can swap them, etc. seems overly complicated for very little gain. Instead we can just emit:

```ts
myMap.forEach(expression);
```

```dart
myMap.forEach((k, v) {
  (expression)(k, k);
});
```

I'd assume that JS VMs are good at unwrapping the immediately invoked function expression, as that's a common pattern. Maybe dart2js even handles that.

vicb:
> Detecting if the passed expression is a function, and if so, if it takes two args, and if so, if we can swap them, etc. seems overly complicated

I actually think that this should be straightforward given an AST.

However it does not seem trivial with TS, may be we can give this feedback to MS if we have privileged contacts there.

mprobst:
Well, the AST is nice and fine, but you'll still end up with some fragile detection code. E.g. think about parenthesized expressions, function literals, arrow fn literals, just passing a reference to a function, passing a reference to a method that'd then be called without a `this` receiver, and so on.

```ts
m.forEach(function(a,b) {...});
m.forEach(function(a, b, c, d) {...});
m.forEach((a,b) => {...});
m.forEach(foo);
m.forEach(x.foo);
m.forEach(y ? ((a, b) => x) : ((a, b) => z));
```

That's not to say it's impossible, and feel free to take a stab at it, but I think it could be a bit tricky. Doing the more generic wrapping into an arrow function might be a good start, and then we could special case specific patterns we recognize starting from there.

WDYT?

vicb:
I think that 99.99% of the calls will be either:
- `m.forEach(function(a, b) { ...});`
- `m.forEach((a, b) => ...);`
- `m.forEach(foo);`
- `m.forEach(x.staticFoo);`

[...]

PR #256, opened 2015-08-04
feat(Facade): add support for Map.forEach

### LINKED ISSUE #253: `Map.forEach` calls the callback with `(key, value)`, but in JavaScript it's `(value, key)`

vicb:
fixes #253
closes #215

@mprobst I would have preferred to factorize some code but not sure if it's worth it due to the subtle diffs of fn vs arrow.

/cc @tbosch 

/ref #227

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -353,6 +353,53 @@ export class FacadeConverter extends base.TranspilerBase {
         this.emitMethodCall('remove', c.arguments);
         this.emit('!= null || true ) )');
       },
+      'Map.forEach': (c: ts.CallExpression, context: ts.Expression) => {
+        let cb: any;
+        let params: any;
+
+        switch (c.arguments[0].kind) {
+          case ts.SyntaxKind.FunctionExpression:
+            cb = <ts.FunctionExpression>(c.arguments[0]);
+            params = cb.parameters;
+            if (params.length != 2) {
+              this.reportError(c, 'Map.forEach callback requires exactly two arguments');
+              return;
+            }
+            this.visit(context);
+            this.emit('. forEach ( (');
+            this.visit(params[1]);
+            this.emit(',');
+            this.visit(params[0]);
+            this.emit(')');
+            this.visit(cb.body);
+            this.emit(')');
+            break;
+
+          case ts.SyntaxKind.ArrowFunction:
+            cb = <ts.ArrowFunction>(c.arguments[0]);
+            params = cb.parameters;
+            if (params.length != 2) {
+              this.reportError(c, 'Map.forEach callback requires exactly two arguments');
+              return;
+            }
+            this.visit(context);
+            this.emit('. forEach ( (');
+            this.visit(params[1]);
+            this.emit(',');
+            this.visit(params[0]);
+            this.emit(') =>');
+            this.visit(cb.body);
+            this.emit(')');
+            break;
+
+          default:
+            this.visit(context);
+            this.emit('. forEach ( ( k , v ) => (');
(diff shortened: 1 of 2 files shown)


lib/facade_converter.ts
```diff
@@ -353,6 +353,50 @@ export class FacadeConverter extends base.TranspilerBase {
...
+      'Map.forEach': (c: ts.CallExpression, context: ts.Expression) => {
+        let cb: any;
+        let params: any;
+
+        switch (c.arguments[0].kind) {
+          case ts.SyntaxKind.FunctionExpression:
+            cb = <ts.FunctionExpression>(c.arguments[0]);
+            params = cb.parameters;
+            if (params.length != 2) {
+              this.reportError(c, 'Map.forEach callback requires exactly two arguments');
```
mprobst: maybe return here? otherwise we'll crash below when passing an undefined `params[1]` down.

vicb: I'll update here and below

lib/facade_converter.ts
```diff
@@ -353,6 +353,50 @@ export class FacadeConverter extends base.TranspilerBase {
...
+            }
+            this.visit(context);
+            this.emit('. forEach ( (');
```
mprobst: shouldn't this have a `function` keyword in here, somewhere?

lib/facade_converter.ts
```diff
@@ -353,6 +353,50 @@ export class FacadeConverter extends base.TranspilerBase {
...
+            this.emit(') =>');
+            this.visit(cb.body);
+            this.emit(')');
+            break;
+
+          default:
+            this.visit(context);
+            this.emit('. forEach ( ( k , v ) => (');
+            this.visit(c.arguments[0]);
+            this.emit(') ( v , k ) )');
```
mprobst: missing `break;`

lib/facade_converter.ts
vicb: in Dart ?

lib/facade_converter.ts
vicb: I'v never used it for `default`, should we ?

lib/facade_converter.ts
vicb: Actually doesn't `reportError()` break the execution flow ? (otherwise there seems to be other places that need to be updated in this file)

vicb: Just tested, it does break the exec flow - I'll add a commit to remove uneeded `return` statements in this file.

vicb:
@mprobst I have updated this PR adding a commit to remove an extra `return`.

I'm not sure if any of the previous comments require some update ? Let me know if you're ok with that.

Thanks for the review.

lib/facade_converter.ts
mprobst: Ah sorry, Dart/JS confusion here ;-)

lib/facade_converter.ts
mprobst: Yes, you could still have cases after default (also a bad practice) and you'd fall through into them, so it's better to just always but break in every.

lib/facade_converter.ts
mprobst: No please don't. It only breaks execution flow during tests so you get a nicer stack trace - but in production use, we want to capture all errors in a list, instead of failing at the first wrong thing (so that a user doesn't have to re-run for every single error).

mprobst:
Thanks Victor, this looks very good with a couple of nits. Please make sure not to remove returns after the `reportError` calls, see comment above. Otherwise this is good to go in.

vicb:
landed (with feedback adressed)

PR #258, opened 2015-08-06
feat(Facade): detect when Array.forEach callback has more than 1 arg

vicb:
Whenever a function (function expression or arrow function) is passed to `Array.forEach()`, makes sure the function takes exactly one parameter (as expected by Dart - JS could have up to 3 params).

Note that no check is performed when the callback is an identifier (`a.forEach(myFn)`).

Supersedes #213

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -283,6 +283,31 @@ export class FacadeConverter extends base.TranspilerBase {
         this.emit(')');
       });
     },
+    'Array.forEach': (c: ts.CallExpression, context: ts.Expression) => {
+      let cb: any;
+      let params: any[];
+
+      switch (c.arguments[0].kind) {
+        case ts.SyntaxKind.FunctionExpression:
+          cb = <ts.FunctionExpression>(c.arguments[0]);
+          params = cb.parameters;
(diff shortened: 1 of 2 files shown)


lib/facade_converter.ts
```diff
@@ -283,6 +283,31 @@ export class FacadeConverter extends base.TranspilerBase {
         this.emit(')');
       });
     },
+    'Array.forEach': (c: ts.CallExpression, context: ts.Expression) => {
```
mprobst: Wouldn't this normally be caught by the Dart analyzer anyway?

vicb: It should.
The only advantage of having the check here it that it would fail earlier.

mprobst: I'm ok with this going in, but overall I think this is not too useful - we don't check other potentially incorrect Dart code either, and of course we cannot in the general case (as we'd need to understand Dart types). So if anything I wouldn't expand on this line of code.

mprobst:
LGTM, though see note.

vicb:
Closing as not really useful (we have to rely on the Dart Analyzer for exhaustive checks anyway)

PR #259, opened 2015-08-07
feat(Facade): Add support for Uint8Array and ArrayBuffer

jteplitz:


lib/facade_converter.ts
```diff
@@ -92,19 +92,33 @@ export class FacadeConverter extends base.TranspilerBase {
...
       }
-      return n.getChildren(sourceFile).some(hasXmlHttpRequest);
-    };
+    }
+  }
 
-    if (hasXmlHttpRequest(sourceFile)) {
-      this.emit('import "dart:html";');
+  private _hasType(n: ts.Node, typeName: String, sourceFile?: ts.SourceFile): boolean {
+    if (n.kind === ts.SyntaxKind.TypeReference) {
```
mprobst: This walks the entire AST for a source file n times for each of the - currently three - types that need special casing. Can we do this in just one walk, e.g. by keeping track of what extra imports we emitted already, and then emitting directly during the search?

jteplitz602: Good idea. Just updated emitImports to do just that.

lib/facade_converter.ts
```diff
@@ -92,19 +95,34 @@ export class FacadeConverter extends base.TranspilerBase {
...
+    var libraries = <ts.Map<string>>{
+      "XMLHttpRequest": "dart:html",
+      "Uint8Array": "dart:typed_arrays",
+      "ArrayBuffer": "dart:typed_arrays"
+    };
+    var emitted: Set = {};
+    this._emitImports(sourceFile, libraries, emitted);
+  }
+
+  private _emitImports(n: ts.Node, libraries: ts.Map<string>, emitted: Set,
```
alexeagle: I don't think we typically use underscore naming in ts2dart (it's pure typescript so the private keyword is sufficient) - can you check that and remove if this is the first underscore prefix being introduced?

lib/facade_converter.ts
```diff
@@ -92,19 +95,34 @@ export class FacadeConverter extends base.TranspilerBase {
...
-        var type = base.ident((<ts.TypeReferenceNode>n).typeName);
-        if (type === 'XMLHttpRequest') {
-          return true;
+    var libraries = <ts.Map<string>>{
+      "XMLHttpRequest": "dart:html",
+      "Uint8Array": "dart:typed_arrays",
+      "ArrayBuffer": "dart:typed_arrays"
+    };
+    var emitted: Set = {};
+    this.emitImports(sourceFile, libraries, emitted);
```
mprobst: I find not passing `sourceFile` in this call and then doing the cast below quite confusing. Could you make the parameter non-optional and just pass `sourceFile` twice here?

lib/facade_converter.ts
```diff
@@ -92,19 +95,34 @@ export class FacadeConverter extends base.TranspilerBase {
...
+    }
 
-    if (hasXmlHttpRequest(sourceFile)) {
-      this.emit('import "dart:html";');
+    if (!sourceFile) {
+      sourceFile = <ts.SourceFile>n;
     }
+
+    n.getChildren(sourceFile)
+        .forEach((n: ts.Node) => { this.emitImports(n, libraries, emitted, sourceFile); });
```
mprobst: I think you can leave off the curly braces here.

test/facade_converter_test.ts
```diff
@@ -73,6 +73,10 @@ describe('type based translation', () => {
       expectWithTypes('var n: Node;').to.equal(' dynamic n ;');
       expectWithTypes('var _xhr: XMLHttpRequest;')
           .to.equal(' import "dart:html"; HttpRequest _xhr ;');
+      expectWithTypes('var _intArray: Uint8Array;')
```
mprobst: could you remove the leading underscores from the variables? Please also remove them on the line above.

mprobst:
@jteplitz602 I noticed this was lying around, going stale. Please remember to comment on the pull request so that your reviewers get notified (something like *"PTAL!"* is enough).

jteplitz602:
Got it. I implemented the changes you suggested. Can you take another look? Also, Travis hasn't kicked in since my latest commit for some reason.

mprobst:
LGTM, please merge.

jteplitz602:
Great. I don't have write access to this repo. Can you merge?

alexeagle:
merged in 124f126

PR #260, opened 2015-08-07
fix(CallTranspiler): handle NewExpressions without parameter lists

caitp:
Previously, ts2dart would crash in the event of an expression like `new
<callee>;` rather than `new <callee>();`.

This minor adjustment improves compat with TypeScript, and is less of a
pain.

### DIFF
--- a/lib/call.ts
+++ b/lib/call.ts
@@ -54,7 +54,7 @@ class CallTranspiler extends base.TranspilerBase {
       this.maybeVisitTypeArguments(c);
     }
     this.emit('(');
-    if (!this.handleNamedParamsCall(c)) {
+    if (c.arguments && !this.handleNamedParamsCall(c)) {
       this.visitList(c.arguments);
     }
     this.emit(')');
--- a/test/call_test.ts
+++ b/test/call_test.ts
@@ -46,4 +46,6 @@ describe('calls', () => {
     expectTranslate('class X { y() { super.z(1); } }')
         .to.equal(' class X { y ( ) { super . z ( 1 ) ; } }');
   });
+  it('transpiles new calls without arguments',
+     () => { expectTranslate('new Foo;').to.equal(' new Foo ( ) ;'); });
 });


caitp:
@alexeagle or someone want to review this very small patch? This crash was annoying to debug in angular2 because there's no information about what causes it.

lib/call.ts
```diff
@@ -26,6 +26,9 @@ class CallTranspiler extends base.TranspilerBase {
           this.emit('new');
         }
         var newExpr = <ts.NewExpression>node;
+        if (newExpr.arguments === undefined) {
```
alexeagle: I think your request is for better error handling, but this silently accepts invalid TS syntax. I think it should just throw with a better error if that's helpful.

test/call_test.ts
```diff
@@ -46,4 +46,6 @@ describe('calls', () => {
     expectTranslate('class X { y() { super.z(1); } }')
         .to.equal(' class X { y ( ) { super . z ( 1 ) ; } }');
   });
+  it('transpiles new calls without arguments',
+     () => { expectTranslate('new Foo;').to.equal(' new Foo ( ) ;'); });
```
alexeagle: expectErroneousCode?

lib/call.ts
caitp: It doesn't look like it is a TS error --- TS generates the right code from it, with no warnings

test/call_test.ts
caitp: I mean, it's perfectly valid code in TypeScript and JS. For some reason the AST just represents it differently from `()`

lib/call.ts
```diff
@@ -26,6 +26,9 @@ class CallTranspiler extends base.TranspilerBase {
           this.emit('new');
         }
         var newExpr = <ts.NewExpression>node;
+        if (!newExpr.arguments) {
+          newExpr.arguments = <ts.NodeArray<ts.Expression>>([]);
```
mprobst: Modifying the TypeScript AST is verboten - TypeScript has internal, unexposed structures that depend on the AST not being modified. Changing properties on it has lead to very, very obscure bugs in the past.

Can you just handle this in `visitCall`?

mprobst:
We might also want to have a linter step that mandates always using parentheses. I don't think there's a good reason to use `new Foo` instead of `new Foo`, except for saving two characters. OTOH this makes it harder to find instantiations using `grep` and friends.

caitp:
> We might also want to have a linter step that mandates always using parentheses. I don't think there's a good reason to use new Foo instead of new Foo, except for saving two characters. OTOH this makes it harder to find instantiations using grep and friends.

http://google.github.io/styleguide/javascriptguide.xml?showone=Parentheses#Parentheses I sort of take this to mean "don't use them" because they aren't required (although you could argue they're needed "semantically") --- if owners want to lint this, I'll change this to lint it, otherwise will change to handle in visitCall

mprobst:
@caitp this is obviously a question of style, so YMMV. But the Google JS style guide is often not a good advisor on issues - in particular removing parentheses from e.g. boolean calculations is often a terrible idea w.r.t. readability.

Nonetheless, `ts2dart` should handle all valid TypeScript, so this is a fine change. And `ts2dart` also isn't a linter, so no change needed for that.

caitp:
1 liner in `CallTranspiler.visitCall()` now.

mprobst:
Pushed as 51b4a5f, thanks for the contribution!

PR #261, opened 2015-08-12
feat: defensive check that a symbol has some declaration.

mprobst:
Possibly due to incorrect TypeScript syntax, symbols might not have any
declarations at all. This adds some defensive code to report that as an error
condition.

The change is a bit larger than expected because we don't have a `ts.Node` down
in that function and need one to give a good location for the error.

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -69,7 +69,7 @@ export class FacadeConverter extends base.TranspilerBase {
       return false;
     }
 
-    var handler = this.getHandler(symbol, this.callHandlers);
+    var handler = this.getHandler(c, symbol, this.callHandlers);
     return handler && !handler(c, context);
   }
 
@@ -83,7 +83,7 @@ export class FacadeConverter extends base.TranspilerBase {
       return false;
     }
 
-    var handler = this.getHandler(symbol, this.propertyHandlers);
+    var handler = this.getHandler(pa, symbol, this.propertyHandlers);
     return handler && !handler(pa);
   }
 
@@ -118,7 +118,7 @@ export class FacadeConverter extends base.TranspilerBase {
         this.reportMissingType(typeName, ident);
         return;
       }
-      let fileAndName = this.getFileAndName(symbol);
+      let fileAndName = this.getFileAndName(typeName, symbol);
       if (fileAndName) {
         var fileSubs = this.TS_TO_DART_TYPENAMES[fileAndName.fileName];
         if (fileSubs && fileSubs.hasOwnProperty(fileAndName.qname)) {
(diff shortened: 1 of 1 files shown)


lib/facade_converter.ts
```diff
@@ -130,19 +130,23 @@ export class FacadeConverter extends base.TranspilerBase {
...
 
-  private getFileAndName(symbol: ts.Symbol): {fileName: string, qname: string} {
+  private getFileAndName(n: ts.Node, symbol: ts.Symbol): {fileName: string, qname: string} {
     while (symbol.flags & ts.SymbolFlags.Alias) symbol = this.tc.getAliasedSymbol(symbol);
     let decl = symbol.valueDeclaration;
     if (!decl) {
       // In the case of a pure declaration with no assignment, there is no value declared.
       // Just grab the first declaration, hoping it is declared once.
+      if (!symbol.declarations || symbol.declarations.length === 0) {
+        this.reportError(n, 'no declarations for symbol');
```
vicb: Should a string representation of the symbol be included in the message ?
I'm actually not sure what the error message would look like when this if branch is taken, would it be possible to add a test ?

mprobst: Done. This will just complain that it cannot find declarations for the given symbol that was encountered at this location. I think that's fair - this shouldn't really happen in the first place (symbol with no declaration doesn't make sense).

lib/facade_converter.ts
```diff
@@ -130,19 +130,23 @@ export class FacadeConverter extends base.TranspilerBase {
...
 
-  private getFileAndName(symbol: ts.Symbol): {fileName: string, qname: string} {
+  private getFileAndName(n: ts.Node, symbol: ts.Symbol): {fileName: string, qname: string} {
     while (symbol.flags & ts.SymbolFlags.Alias) symbol = this.tc.getAliasedSymbol(symbol);
     let decl = symbol.valueDeclaration;
     if (!decl) {
       // In the case of a pure declaration with no assignment, there is no value declared.
       // Just grab the first declaration, hoping it is declared once.
+      if (!symbol.declarations || symbol.declarations.length === 0) {
+        this.reportError(n, 'no declarations for symbol ' + symbol.name);
```
vicb: `no declarations for symbol ${symbol.name}` ?
This looks so ES5 :)

vicb:
LGTM

lib/facade_converter.ts
mprobst: But IMHO that's really just adding syntactical cruft here. I like the interpolation for when I'm actually interpolating, but for plain concatenation this is IMHO better.

PR #264, opened 2015-08-19
fix: use `implements` for interface extension.

### LINKED ISSUE #263: Interfaces that extend other interfacesin TS should implement them in dart
Currently the following TypeScript code

```TypeScript
interface A {...}

interface B extends A {...}
```
Gets transpiled to the following Dart
```dart
abstract class A {...}
abstract class B extends A {...}
```
But we should really be using implements instead of extends here to more accurately emulate what's going in TypeScript.
For example, with extends in Dart it is not possible for a class to extend multiple parent classes but in TypeScript an interface can implement multiple parent interfaces. Currently,
```TypeScript
interface A {...}

interface B {...}

interface A extends A, B {...}
```
becomes
```dart
abstract class A {...}
abstract class B {...}

abstract class C extends A, B {...}
```
Which is not valid Dart, but

```dart
abstract class A {...}
abstract class B {...}

abstract class C implements A, B {...}
```
Would be correct.

mprobst:
TypeScript interfaces are represented as abstract classes in Dart. Abstract
classes can either implement or extend super classes, but an interface extending
multiple super interfaces should be represented by `implements`, as Dart does
not have multiple inheritance for abstract classes.

Fixes #263.

### DIFF
--- a/test/declaration_test.ts
+++ b/test/declaration_test.ts
@@ -102,23 +102,27 @@ describe('classes', () => {
       expectTranslate('class X { constructor() { } }').to.equal(' class X { X ( ) { } }');
     });
     it('supports parameter properties', () => {
+      expectTranslate('class X { c: number; \n' +
+                      '  constructor(private bar: B, ' +
+                      'public foo: string = "hello", ' +
+                      'protected goggles: boolean = true) {} }')
+          .to.equal(' class X { B bar ; String foo ; bool goggles ; num c ;' +
+                    ' X ( this . bar , [ this . foo = \"hello\" , this . goggles = true ] ) { } }');
       expectTranslate(
-          'class X { c: number; constructor(private bar: B, public foo: string = "hello", protected goggles: boolean = true) {} }')
-          .to.equal(
-              ' class X { B bar ; String foo ; bool goggles ; num c ; X ( this . bar , [ this . foo = \"hello\" , this . goggles = true ] ) { } }');
-      expectTranslate(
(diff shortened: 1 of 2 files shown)


test/declaration_test.ts
```diff
@@ -102,23 +102,27 @@ describe('classes', () => {
       expectTranslate('class X { constructor() { } }').to.equal(' class X { X ( ) { } }');
     });
     it('supports parameter properties', () => {
+      expectTranslate('class X { c: number; \n' +
```
mprobst: FYI these changes are unrelated, I just wanted to fix the lines >> 100 chars here.

test/declaration_test.ts
```diff
@@ -102,23 +102,27 @@ describe('classes', () => {
...
-  it('should translate interfaces',
+  it('translates interfaces to abstract classes',
      () => { expectTranslate('interface X {}').to.equal(' abstract class X { }'); });
-  it('should support extends', () => {
-    expectTranslate('interface X extends Y, Z {}').to.equal(' abstract class X extends Y , Z { }');
+  it('translates interface extends to class implements', () => {
+    expectTranslate('interface X extends Y, Z {}')
+        .to.equal(' abstract class X implements Y , Z { }');
   });
   it('should support implements', () => {
```
vicb: _Not directly related to this PR:_
Seems like the 2 following tests are duplicates (from the 'classes' group)

vicb:
LGTM with one comment about duplicate tests (not directly related to this PR)

test/declaration_test.ts
mprobst: Good catch, done.

PR #266, opened 2015-08-24
chore: update facade paths to new paths in core

jeffbcross:
As part of angular/angular#3713, facade lives in the core directory
inside angular2.


jeffbcross:
Per @alexeagle 's suggestion, I should parameratize this.

jeffbcross:
I should probably update tests, too.

alexeagle:
I don't have a good suggestion how angular would pass this map through to ts2dart. I suppose something like a json config file we check in, and hand the path to that file when brocolli-ts2dart invokes ts2dart in-process?

alexeagle:
LGTM on the change, when you've updated tests as needed. Parameterizing might be out of scope for this change, up to you

jeffbcross:
@alexeagle how about adding a property to the options argument passed to ts2dart. 

```typescript
ts2dart(tsInputTree, {
  facadeBasePath: 'angular2/src/core/facade'
  ...
});
```

alexeagle:
If we are going to add the parameter, I think we should completely decouple angular's type mapping from ts2dart. It's not worth adding the ability to change facadeBasePath without also adding the ability to arbitrarily change the type mapping.

alexeagle:
Oh and that would work well for the first block of type replacements, but
not so well for the call replacements.

jeffbcross:
@alexeagle good point. In that case, I will forego parameretizing for now and will open an issue to explore decoupling.

PR #269, opened 2015-08-25
change di path

jeffbcross:
Guess I also forgot to push the package.json bump.


jeffbcross:
This was supposed to be done in #266 but I missed it. TL;DR facade, di, and some other paths are being moved to inside of core. See angular/angular#3713

alexeagle:
LGTM

PR #279, opened 2015-09-04
feat: reenable check for visibility modifiers

### LINKED ISSUE #278: Disallow private fields without leading _
Time to fix this old TODO: https://github.com/angular/ts2dart/blob/6f3bbd896e7e712049eb4f2403e67f59932cf697/lib/declaration.ts#L426

harryterkelsen:
Fixes #278

### DIFF
--- a/lib/declaration.ts
+++ b/lib/declaration.ts
@@ -156,6 +156,7 @@ class DeclarationTranspiler extends base.TranspilerBase {
         if (this.hasFlag(paramDecl.modifiers, ts.NodeFlags.Public) ||
             this.hasFlag(paramDecl.modifiers, ts.NodeFlags.Private) ||
             this.hasFlag(paramDecl.modifiers, ts.NodeFlags.Protected)) {
+          this.visitDeclarationMetadata(paramDecl);
           this.emit('this .');
           this.visit(paramDecl.name);
           if (paramDecl.initializer) {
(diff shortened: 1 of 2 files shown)


lib/declaration.ts
```diff
@@ -422,13 +426,11 @@ class DeclarationTranspiler extends base.TranspilerBase {
...
     this.visitEachIfPresent(decl.modifiers);
 
-    // Temporarily deactivated to make migration of Angular code base easier.
-    return;
-
     if (this.hasFlag(decl.modifiers, ts.NodeFlags.Protected)) {
       this.reportError(decl, 'protected declarations are unsupported');
       return;
     }
+    if (!decl.name) return;
```
mprobst: Could you add a comment explaining when this happens?

hterkelsen: I don't remember what was causing it when I put the check in, but I took it out again and the tests all pass, so I'll just leave it out

hterkelsen: Nevermind. It happens with constructors. Added comment

mprobst:
LGTM, please merge.

hterkelsen:
Should I merge even if there are several hundred places in Angular that will become errors if this check is turned on?

mprobst:
Yes, but that makes you responsible to fix those in Angular when you upgrade ts2dart :-)

hterkelsen:
I've already fixed all of the private fields without leading underscores. I would actually like to allow public fields with leading underscores since that is helpful for dart2js. What do you think?

mprobst:
Merged as e288c35.

As discussed offline, let's make all those public fields leading with an underscore in angular `private`, as they should be.

PR #280, opened 2015-09-08
feat(Array): add support for join

vicb:
@mprobst could you please review

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -303,6 +303,14 @@ export class FacadeConverter extends base.TranspilerBase {
       });
       this.emit(')');
     },
+    'Array.join': (c: ts.CallExpression, context: ts.Expression) => {
+      this.visit(context);
+      if (c.arguments.length) {
+        this.emitMethodCall('join', c.arguments);
+      } else {
+        this.emit('. join ( "," )');
+      }
+    },
     'ArrayConstructor.isArray': (c: ts.CallExpression, context: ts.Expression) => {
       this.emit('( (');
       this.visitList(c.arguments);  // Should only be 1.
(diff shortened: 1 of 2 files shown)


lib/facade_converter.ts
```diff
@@ -303,6 +303,14 @@ export class FacadeConverter extends base.TranspilerBase {
       });
       this.emit(')');
     },
+    'Array.join': (c: ts.CallExpression, context: ts.Expression) => {
+      this.visit(context);
+      if (c.arguments.length === 0) {
```
mprobst: nit: I'm ok with writing `if (c.arguments.length) { ... }`, that's idiomatic enough in my opinion. YMMV.

test/facade_converter_test.ts
```diff
@@ -94,32 +94,46 @@ describe('type based translation', () => {
     it('translates array operations to dartisms', () => {
       expectWithTypes('var x: Array<number> = []; x.push(1); x.pop();')
           .to.equal(' List < num > x = [ ] ; x . add ( 1 ) ; x . removeLast ( ) ;');
+
```
mprobst: Why the newlines here, but not below? I think I'd rather not have many newlines, but group related blocks of tests within this `it` by a newline.

mprobst:
LGTM.

lib/facade_converter.ts
vicb: right, we don't need to transpile the transpiler, I'll update

test/facade_converter_test.ts
vicb: I'll remove the newlines from here (indeed, the idea was to group related tests which explain why there is no newlines between `unshift` and `unshift`)

vicb:
landed as b642cf10d01afed7f7b67e93c0d936cdf6fdcc84

PR #284, opened 2015-09-23
feature(options): allow configuration of enforcing private/public und…

alexeagle:
…erscore convention

angular is not yet updated to comply with the new rules


IgorMinar:
looks like docs are missing. is there a good place to document the options?

IgorMinar:
otherwise lgtm

alexeagle:
Yes we don't document ts2dart very well. We should add this to the cheatsheet. Merging for now.

PR #286, opened 2015-09-25
chore(typings): list the external API definition in package.json.

alexeagle:
This is supported by new Typescript ModuleResolutionKind=node as described in the proposal here:
https://github.com/Microsoft/TypeScript/issues/2338


rkirov:
LGTM

PR #287, opened 2015-09-25
Typings

alexeagle:


rkirov:
Ops, one more version. No big deal, looks like a very active project. LGTM

PR #293, opened 2015-10-02
feature: support index signatures

alexeagle:
supports https://github.com/angular/angular/issues/4483

### DIFF
--- a/test/type_test.ts
+++ b/test/type_test.ts
@@ -6,6 +6,11 @@ describe('types', () => {
      () => { expectTranslate('var x: foo.Bar;').to.equal(' foo . Bar x ;'); });
   it('drops type literals',
      () => { expectTranslate('var x: {x: string, y: number};').to.equal(' dynamic x ;'); });
+  it('translates string index signatures to dartisms', () => {
+    expectTranslate('var x: {[k: string]: any[]};')
+        .to.equal(' Map < String , List < dynamic > > x ;');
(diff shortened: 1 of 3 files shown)


alexeagle:
merged as e16c3543070fda6db831638606dbf23ac7c81c4f

test/type_test.ts
```diff
@@ -6,6 +6,11 @@ describe('types', () => {
      () => { expectTranslate('var x: foo.Bar;').to.equal(' foo . Bar x ;'); });
   it('drops type literals',
      () => { expectTranslate('var x: {x: string, y: number};').to.equal(' dynamic x ;'); });
+  it('translates string index signatures to dartisms', () => {
+    expectTranslate('var x: {[k: string]: any[]};')
```
vicb: @alexeagle Would this PR allow `'var x: {size: number, [k: string]: any[]};'` ?

We should have test for that.

alexeagle: Well, that particular example isn't not allowed because all the properties
have to have the type that the indexer suggests. In fact, that's a useful
pattern, you can specify part of the object shape and allow "other" as well
using
`{size: number, [key: string]: number, height: number, width: number}'
and that matches {size: 3, height: 4, width: 5, depth: 6}

However, we would not be able to type such a thing in Dart, so you're right
that this PR is missing a constraint. I'll add that.

PR #294, opened 2015-10-03
fix(type literals): detect index signatures with other properties

alexeagle:


lib/type.ts
```diff
@@ -9,27 +9,21 @@ class TypeTranspiler extends base.TranspilerBase {
...
-                this.emit(',');
-                this.visit(indexSig.type);
-                this.emit('>');
-              }
-            });
-        if (!foundIndexSignature) {
-          // Dart doesn't support other type literals.
-          this.emit('dynamic');
+        let members = (<ts.TypeLiteralNode>node).members;
+        if (members.length == 1 && members[0].kind == ts.SyntaxKind.IndexSignature) {
```
vicb: What about `{name: string, [extra: string]: string}` ie all of the same type even if `members.length > 1`

vicb:
Thanks for the fix, we can even type more cases in Dart, see inline comments

lib/type.ts
alexeagle: What Dart would we produce for that? And do you imagine a place where we would want to use it?

vicb: Dart would produce a string to string map. But you're right, we would loose some info in the process, not sure if it would be valuable.

That's a pattern that could be used when you care only about some of the keys but others exist.

mprobst: You could generate a subtype of map that adds the properties. But that risks optimization losses (subtypes of Map and Array are dangerous), and overall seems too clever for its own good. Let's be stupid here.

alexeagle: okay, so LG as is then?

vicb: TM

PR #295, opened 2015-10-07
feat(Array): add support for some, find & filter

vicb:
@alexeagle @mprobst I would need some help with this one.

The tests currently fails with "access to untyped property..."
If I add the Array interface with the find method as this is done for the the Map, the error disappears but `find` is not translated to `firstWhere`.

### DIFF
--- a/test/facade_converter_test.ts
+++ b/test/facade_converter_test.ts
@@ -13,6 +13,9 @@ var es6RuntimeDeclarations = `
       delete(key: K): boolean;
       forEach(callbackfn: (value: V, index: K, map: Map<K, V>) => void, thisArg?: any): void;
     }
+    interface Array<T> {
+      find(predicate: (value: T, index: number, obj: Array<T>) => boolean, thisArg?: any): T;
+    }
     declare var Map: {
       new<K, V>(): Map<any, any>;
       prototype: Map<any, any>;
@@ -97,6 +100,8 @@ describe('type based translation', () => {
           .to.equal(' List < num > x = [ ] ; x . add ( 1 ) ; x . removeLast ( ) ;');
       expectWithTypes('var x: Array<number> = []; x.map((e) => e);')
           .to.equal(' List < num > x = [ ] ; x . map ( ( e ) => e ) . toList ( ) ;');
+      expectWithTypes('var x: Array<number> = []; x.filter((e) => true);')
+          .to.equal(' List < num > x = [ ] ; x . where ( ( e ) => true ) . toList ( ) ;');
       expectWithTypes('var x: Array<number> = []; x.unshift(1, 2, 3); x.shift();')
           .to.equal(' List < num > x = [ ] ; ( x .. insertAll ' +
                     '( 0, [ 1 , 2 , 3 ]) ) . length ; x . removeAt ( 0 ) ;');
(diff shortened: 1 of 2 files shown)


test/test_support.ts
```diff
@@ -7,7 +7,9 @@ import main = require('../lib/main');
 import path = require('path');
 import ts = require('typescript');
 
-export type StringMap = { [k: string]: string };
+export type StringMap = {
```
alexeagle: can we remove this now? Sorry this is probably my leftover bit

vicb: let's do this as part of an other PR. I haven't touch this code (it only has been automatically reformated)

mprobst:
I presume this doesn't work because:
- `Array.find` is an ES6 thing, so it's not just available on `Array.prototype`
- When you add it like the `Map` stuff in the test, it's not defined in `lib.d.ts`

If these are ES6-isms, do we actually have a shim/polyfill that implements them for Angular? Where is that defined? Should we just use those definitions as the location?

vicb:
@mprobst could you please check the latest updates (thanks for the tips).

vicb:
Hum interesting...

I had to revert to clang-format 1.0.29 locally because Travis would fail (I had 1.0.30 installed) on this PR.

- This PR used to pass earlier with the same formatting on this repo,
- My local repo (vicb)  [fails with the current code](https://travis-ci.org/vicb/ts2dart/builds/84030087#L197) - it uses 1.0.30

Any idea ?

mprobst:
Code LGTMs.

If you want to use a newer clang-format, you have to update the NPM version; clang-format should normally use the version that is installed to `./node_modules`. I guess we should upgrade, we had some useful changes in clang-format.

vicb:
I will check this version within Angular before merging.

vicb:
Tested with Angular, works fine.

/cc @mprobst @alexeagle

test/facade_converter_test.ts
```diff
@@ -113,6 +118,11 @@ describe('type based translation', () => {
           .to.equal(' List < num > x = [ ] ; String y = x . join ( "-" ) ;');
       expectWithTypes('var x: Array<number> = []; var y: string = x.join();')
           .to.equal(' List < num > x = [ ] ; String y = x . join ( "," ) ;');
+      expectWithTypes('var x: Array<number> = []; var y: number = x.find((e) => e == 0);')
+          .to.equal(
+              ' List < num > x = [ ] ; num y = x . firstWhere ( ( e ) => e == 0 , orElse : ( ) => null ) ;');
```
alexeagle: these tests are kind of lame. It would be easy to assert we get a certain dart output that doesn't actually work. Perhaps we could do some testing in the e2e tests as well, though on the other hand we don't want to exhaustively cover the cases in that style test.

So clearly I don't have very good guidance here, this is certainly fine for the present PR.

alexeagle:
Go ahead and merge (this repo doesn't have any presubmit fanciness)

vicb:
@alexeagle I agree about the test and I think this has caused issues in the past. I have tested this particular PR within angular so I do not expect issue but I agree we could & should do a better job.

I'll amend the commit with `Array.reduce` & merge. Thanks for the review.

PR #299, opened 2015-10-08
chore(stringmap): remove StringMap support which is removed from angular

alexeagle:


mprobst:
LGTM.

alexeagle:
merged as dacca0c

PR #300, opened 2015-10-09
Moar dart html

alexeagle:
Fixes the build failure of https://github.com/angular/angular/pull/4636

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -95,6 +95,7 @@ export class FacadeConverter extends base.TranspilerBase {
   emitExtraImports(sourceFile: ts.SourceFile) {
     var libraries = <ts.Map<string>>{
       "XMLHttpRequest": "dart:html",
+      "KeyboardEvent": "dart:html",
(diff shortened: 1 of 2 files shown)


lib/facade_converter.ts
```diff
@@ -95,6 +95,7 @@ export class FacadeConverter extends base.TranspilerBase {
   emitExtraImports(sourceFile: ts.SourceFile) {
     var libraries = <ts.Map<string>>{
       "XMLHttpRequest": "dart:html",
+      "KeyboardEvent": "dart:html",
```
mprobst: What about `MouseEvent`? Could we add some more of the usual suspects here?

alexeagle: Nope, I tried that but we cast them to dynamic and then they are erased,
and we get an error from dartanalyze because the dart:html import goes
unused.

vicb:
Please ping me when Angular uses 0.7.12 so that I can rebase my Facade PR. Thanks.

alexeagle:
I have added that to my pending change

PR #301, opened 2015-10-21
chore: clean up gulpfile, use tsconfig.json.

mprobst:
- Always use tsconfig.json as the one authoritative source for
  compilation.
- Specify noEmit so that IDEs don't generate files on disk.
- Override tsconfig.json options as needed:
  - noEmit: false to emit files
  - declaration: true to generate .d.ts files
  - typescript to specify the TS version used


mprobst:
@alexeagle says LGTM.

PR #303, opened 2015-11-04
Remove fs-extra dependency from main.ts

vikerman:
Remove fs-extra from the main package dependencies to avoid pulling in all the packages fs-extra depends on.

Move fs-extra to devDependencies as it's still needed for gulp tasks.

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -1,7 +1,6 @@
 require('source-map-support').install();
 import SourceMap = require('source-map');
 import fs = require('fs');
-import fsx = require('fs-extra');
 import path = require('path');
 import ts = require('typescript');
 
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,6 +16,7 @@
   "devDependencies": {
     "chai": "^2.1.1",
     "clang-format": "^1.0.25",
+    "fs-extra": "^0.18.0",
```
alexeagle: do we actually use this at development time?

lib/main.ts
```diff
@@ -174,6 +173,39 @@ export class Transpiler {
     return this.output.getResult();
   }
 
+  private recursiveMkdirSync(p: string) {
```
alexeagle: should be static, since it does not reference any class members in the implementation

alexeagle:
LGTM with that one fix

vikerman:
Done.

alexeagle:
merged in 0ac845054226cea24af60ef2045e0a373c854c87

PR #310, opened 2015-12-03
Fixes to t2dart to make it run within Google

vikerman:
Upstreaming local Google changes to public github repository.
- ts2dart should now handle relative path names as input and basePath. Fixed tests accordingly.
- Removed inlined fs-extra code and switched to using our own mkdirP implementation.
- Also fixed some bugs with mkdirP.

### DIFF
--- a/lib/mkdirp.ts
+++ b/lib/mkdirp.ts
@@ -1,14 +1,17 @@
 import * as fs from 'fs';
 import * as path from 'path';
 
-
 export default function mkdirP(p: string) {
-  let absPath = process.cwd();
+  // Convert input path to absolute and then relative so that we always have relative path in the
+  // end. This can be made simpler when path.isAbsolute is available in node v0.12.
+  p = path.resolve(p);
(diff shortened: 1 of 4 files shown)


vikerman:
I changed to use basePath without the cwd. Also added test case to cover all combinations of the input file paths and destination directory being relative and absolute.

Changed mkdirP to the way you suggested.

mprobst:
@rkirov could you review this? I'm stuck in an infinite meeting.

lib/mkdirp.ts
```diff
@@ -1,14 +1,15 @@
 import * as fs from 'fs';
 import * as path from 'path';
 
-
 export default function mkdirP(p: string) {
-  let absPath = process.cwd();
+  p = path.resolve(p);
```
rkirov: add a comment, what is this achieving.

rkirov:
looks good, just explain the resolve/relative dance.

PR #311, opened 2015-12-04
feat(module): Add an option to specify a library name prefix

vikerman:
Add a new option libraryNamePrefix which will add a prefix to the generated library names.

This is useful when the library name generated by basePath is removing some part of the library name that's needed.

Ex. For Angular 2 - basePath=modules/angular2, libraryNamePrefix=angular2 would still generate the right library names with "angular2.xxx.xxx"

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -23,6 +23,11 @@ export interface TranspilerOptions {
   failFast?: boolean;
   /** Whether to generate 'library a.b.c;' names from relative file paths. */
   generateLibraryName?: boolean;
+
+  /** A prefix to be prepended to all library names. This is useful when specifying the base path
+   * generates the correct destination file name but the generated library name is not sufficient.
+   */
+  libraryNamePrefix?: string;
   /** Whether to generate source maps. */
   generateSourceMap?: boolean;
   /**
@@ -71,7 +76,7 @@ export class Transpiler {
       new DeclarationTranspiler(this, this.fc, options.enforceUnderscoreConventions),
       new ExpressionTranspiler(this, this.fc),
       new LiteralTranspiler(this, this.fc),
-      new ModuleTranspiler(this, this.fc, options.generateLibraryName),
+      new ModuleTranspiler(this, this.fc, options.generateLibraryName, options.libraryNamePrefix),
       new StatementTranspiler(this),
       new TypeTranspiler(this, this.fc),
     ];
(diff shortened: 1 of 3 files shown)


lib/main.ts
```diff
@@ -23,6 +23,10 @@ export interface TranspilerOptions {
   failFast?: boolean;
   /** Whether to generate 'library a.b.c;' names from relative file paths. */
   generateLibraryName?: boolean;
+  /*** A prefix to be prepended to all library names. This is useful when specifying the base path
```
rkirov: follow `/**\n` convention here

test/module_test.ts
```diff
@@ -62,23 +62,33 @@ describe('library name', () => {
   var transpiler: main.Transpiler;
   var modTranspiler: ModuleTranspiler;
   beforeEach(() => {
-    cwd = process.cwd();
-    transpiler =
-        new main.Transpiler({failFast: true, generateLibraryName: true, basePath: cwd + '/a'});
-    modTranspiler = new ModuleTranspiler(transpiler, new FacadeConverter(transpiler), true);
+    transpiler = new main.Transpiler({failFast: true, generateLibraryName: true, basePath: 'a'});
+    modTranspiler = new ModuleTranspiler(transpiler, new FacadeConverter(transpiler), true, null);
```
rkirov: libraryNamePrefix is optional so just skip passing it instead of passing null

test/module_test.ts
```diff
@@ -62,23 +62,33 @@ describe('library name', () => {
...
   });
   it('adds a library name', () => {
     var results = translateSources(
-        {'a/b/c.ts': 'var x;'}, {failFast: true, generateLibraryName: true, basePath: cwd + '/a'});
+        {'a/b/c.ts': 'var x;'}, {failFast: true, generateLibraryName: true, basePath: 'a'});
```
rkirov: loooks like left-over from previous change. Is it needed?

lib/main.ts
vikerman: Done.

test/module_test.ts
vikerman: Done.

test/module_test.ts
vikerman: I had missed cleaning it up in the last change. Cleaned it up while I add some tests for this change.

lib/main.ts
```diff
@@ -23,6 +23,10 @@ export interface TranspilerOptions {
   failFast?: boolean;
   /** Whether to generate 'library a.b.c;' names from relative file paths. */
   generateLibraryName?: boolean;
+  /** A prefix to be prepended to all library names. This is useful when specifying the base path
```
mprobst: Nit: line break before

mprobst:
LGTM

PR #313, opened 2015-12-09
chore(build): update to typescript 1.7.3

alexeagle:


alexeagle:
Angular is waiting on this, and it's trivial (no changes in ts2dart seem needed) so I'll merge.

mprobst:
LGTM. Sorry, I was on a train to New Jersey.

mprobst:
... should we maybe make this a peer dependency?

alexeagle:
Yeah I don't really understand the changes in peer dependencies coming in
NPM 3 but indeed we would like to use the version of typescript brought by
the user rather than our own.

Issue #307, opened 2015-11-13
basePath Transpiler option is applied to generated library names only if the input file paths are absolute


vikerman:
I am using ts2dart in a build environment to transpile Angular 2 where the input paths are relative but I still want basePath to be applied.

Would it be ok to change basePath to be applied even if the paths are relative? It's not clear why this restriction exists.

PR #315, opened 2015-12-09
fix(angular): don't rewrite relative paths to lib.d.ts

### LINKED ISSUE #314: relative path to stdlib changed
When we run latest ts2dart from angular, we don't convert the facade methods.
The facade_converter has a map from filename to the types we understand, one such filename is "lib"
with the current ts2dart we instead see "../../lib".
Looks like this was introduced in https://github.com/angular/ts2dart/commit/4ae749e597303b7606d647b23daed1463943bc59#diff-dfc2e8680f99b78f2e7d8c13b9a44855L249

I'm going to restore that conditional check to un-block angular. @vikerman this might break the g3 usage?

alexeagle:
Fixes #314

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -5,7 +5,6 @@ import path = require('path');
 import ts = require('typescript');
 
 import base = require('./base');
-import mkdirP from './mkdirp';
 import CallTranspiler = require('./call');
 import DeclarationTranspiler = require('./declaration');
 import ExpressionTranspiler = require('./expression');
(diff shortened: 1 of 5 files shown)


alexeagle:
maybe this answers the question from #307

IgorMinar:
lgtm

IgorMinar:
we'll need to clean this up and bring back the changes after the next release. now we need to unblock https://github.com/angular/angular/pull/5758

alexeagle:
Reverted two commits including cb6d27a51ba5cf99d50ff8b205fd8c9e1e0b68d4

PR #316, opened 2015-12-11
feat: support loading rxjs from node_modules.

mprobst:
Angular no longer declares Observable itself, but rather re-exports from
rxjs/Observable. This change adjusts to the new location, and also
supports loading from node_modules.

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -145,21 +145,24 @@ export class FacadeConverter extends base.TranspilerBase {
   }
 
   private getHandler<T>(n: ts.Node, symbol: ts.Symbol, m: ts.Map<ts.Map<T>>): T {
-    var {fileName, qname} = this.getFileAndName(n, symbol);
+    var loc = this.getFileAndName(n, symbol);
+    if (!loc) return null;
+    var {fileName, qname} = loc;
     var fileSubs = m[fileName];
     if (!fileSubs) return null;
     return fileSubs[qname];
   }
 
-  private getFileAndName(n: ts.Node, symbol: ts.Symbol): {fileName: string, qname: string} {
+  private getFileAndName(n: ts.Node, originalSymbol: ts.Symbol): {fileName: string, qname: string} {
+    let symbol = originalSymbol;
     while (symbol.flags & ts.SymbolFlags.Alias) symbol = this.tc.getAliasedSymbol(symbol);
     let decl = symbol.valueDeclaration;
     if (!decl) {
       // In the case of a pure declaration with no assignment, there is no value declared.
       // Just grab the first declaration, hoping it is declared once.
       if (!symbol.declarations || symbol.declarations.length === 0) {
-        this.reportError(n, 'no declarations for symbol ' + symbol.name);
-        return;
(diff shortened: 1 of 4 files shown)


lib/facade_converter.ts
```diff
@@ -246,8 +246,7 @@ export class FacadeConverter extends base.TranspilerBase {
     'lib.es6': this.stdlibTypeReplacements,
     'angular2/typings/es6-promise/es6-promise': {'Promise': 'Future'},
     'angular2/typings/es6-shim/es6-shim': {'Promise': 'Future'},
-    'angular2/src/facade/async':
-        {'Observable': 'Stream', 'ObservableController': 'StreamController'},
+    'rx/Observable': {'Observable': 'Stream'},
```
vsavkin: This should be `rxjs/Observable`.  

But it is actually worse than that. It looks like because we do not use the node resolution strategy, typescript makes the path to rxjs/Observable relative. So the working configuration looks like this:

```
'../../../../../../rxjs/Observable': { 'Observable': 'Stream' },
```

lib/main.ts
```diff
@@ -139,16 +139,24 @@ export class Transpiler {
...
+          contents = fs.readFileSync(nodePath, 'utf-8');
         }
-        if (!fs.existsSync(path)) return undefined;
-        var contents = fs.readFileSync(path, 'UTF-8');
         return ts.createSourceFile(sourceName, contents, COMPILER_OPTIONS.target, true);
       },
       writeFile(name, text, writeByteOrderMark) { fs.writeFile(name, text); },
-      fileExists: (filename) => fs.existsSync(filename),
+      fileExists: (filename) => {
+        return !!filename.match(/^rxjs\/.*\.d\.ts$/) || fs.existsSync(filename);
```
vsavkin: Change to: `return !!filename.match(/^(\/)*rxjs\/.*\.d\.ts$/) || fs.existsSync(filename);`

This is because the path is '/rxjs/Observable', not 'rxjs/Observable'.

lib/main.ts
```diff
@@ -139,16 +139,24 @@ export class Transpiler {
     defaultLibFileName = this.normalizeSlashes(defaultLibFileName);
     return {
       getSourceFile: (sourceName, languageVersion) => {
-        var path = sourceName;
+        var sourcePath = sourceName;
```
vsavkin: The working version:
```
                var sourcePath = sourceName;
                if (sourceName === defaultLibFileName) {
                    sourcePath = ts.getDefaultLibFilePath(exports.COMPILER_OPTIONS);
                }
                var contents;
                if (fs.existsSync(sourcePath)) {
                    contents = fs.readFileSync(sourcePath, 'UTF-8');
                } else {
                    var nodePath = path.resolve('node_modules', sourceName.substring(1));
                    if (fs.existsSync(nodePath)) {
                        contents = fs.readFileSync(nodePath, 'utf-8');
                    }
                }
                if (contents) {
                    return ts.createSourceFile(sourceName, contents, exports.COMPILER_OPTIONS.target, true);
                } else {
                    return undefined;
                }
```

PR #317, opened 2015-12-12
feat: support loading rxjs from node_modules.

vsavkin:
Angular no longer declares Observable itself, but rather re-exports from
rxjs/Observable. This change adjusts to the new location, and also
supports loading from node_modules.


vicb:
missing BREAKING CHANGE in the commit comment ?

alexeagle:
Is this stale?

mprobst:
I think it is, I fixed it separately.

PR #320, opened 2015-12-19
feat(filesystem): Apply basePath to relative input TS files and library names

vikerman:
Doesn't change module file name resolution

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -80,7 +80,7 @@ export class Transpiler {
    */
   transpile(fileNames: string[], destination?: string): void {
     if (this.options.basePath) {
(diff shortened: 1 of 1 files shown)


vikerman:
Tested with angular/angular that the code generated with and without the change are the same.

alexeagle:
This is okay with me, but it really needs some tests that fail before the change. Can we find the right fixture for a test that reproduces the path problems? I'm imagining a test that lays out a temp directory that looks similar to the google3 scenario, and maybe another that's layed out like angular/angular, and asserts that the program can run on trivially simple inputs.

We could do that in a follow-up change so we're not blocked on google3/ts2dart.

alexeagle:
LGTM

Issue #321, opened 2016-01-09
TypeError in ModuleTranspiler.isEmptyImport on bare import


jsayol:
Having a simple `foo.ts` with
```typescript
import 'bar'
```

Running `node ts2dart/build/lib/main.js foo.ts --destination out` throws a TypeError:

```
Transpiling [ 'foo.ts' ] to out

/source/module.ts:131
    var bindings = n.importClause.namedBindings;
                                 ^
TypeError: Cannot read property 'namedBindings' of undefined
    at ModuleTranspiler.isEmptyImport (/source/module.ts:131:34)
    at ModuleTranspiler.visitNode (/source/module.ts:27:18)
    at Transpiler.visit (/source/main.ts:257:31)
    at ModuleTranspiler.TranspilerBase.visit (/source/base.ts:19:39)
    at visitEachNode (/home/josep/projects/ionic-dart/ts2dart/node_modules/typescript/lib/typescript.js:7421:30)
    at Object.forEachChild (/home/josep/projects/ionic-dart/ts2dart/node_modules/typescript/lib/typescript.js:7578:24)
    at ModuleTranspiler.visitNode (/source/module.ts:20:12)
    at Transpiler.visit (/source/main.ts:257:31)
    at Transpiler.translate (/source/main.ts:175:10)
    at /source/main.ts:107:31
```

jsayol:
While bare imports are unsupported, that check should be done before isEmptyImport is called.

PR #322, opened 2016-01-09
Avoid TypeError when encountering a bare import

jsayol:
See issue #321


mprobst:
Thanks!

Merged in 1cd2817.

PR #325, opened 2016-01-20
Upgrade TypeScript dependency to 1.8-beta (including baseurl resolution)

mprobst:
This is WIP and doesn't quite compile yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "minimist": "^1.1.1",
     "source-map": "^0.4.2",
     "source-map-support": "^0.3.1",
-    "typescript": "1.7.3"
+    "typescript": "mprobst/TypeScript#pathMappingModuleResolution"
   },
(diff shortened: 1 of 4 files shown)


mprobst:
@vikerman PTAL, this should be ready now, thanks to the help of @vladima.

package.json
```diff
@@ -11,7 +11,7 @@
     "minimist": "^1.1.1",
     "source-map": "^0.4.2",
     "source-map-support": "^0.3.1",
-    "typescript": "1.7.3"
+    "typescript": "mprobst/TypeScript#pathMappingModuleResolution"
```
vikerman: Should be 1.8?

mprobst: This matches the version we're using internally at Google. We should upgrade to 1.8 as soon as that is released.

vikerman:
LGTM

PR #326, opened 2016-01-20
fix breaking changes in API introduces in TS-1.8

vladima:
added a few upcases that were previously unnecessary because the same result was obtained via subtype reduction


mprobst:
Thanks @vladima! I'll close this PR in favour of the other one that also upgrades to TS 1.8 (beta) and includes some more API fixes/adjustments.

PR #327, opened 2016-01-27
Integrate Dart Formatter directly into ts2dart for cleaner output

jacob314:
and more readable test cases.

Not sure if the way I integrated dart-style.d.ts is the right but it works.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,6 +12,7 @@ var sourcemaps = require('gulp-sourcemaps');
 var spawn = require('child_process').spawn;
 var ts = require('gulp-typescript');
 var typescript = require('typescript');
+var style = require('dart-style');
 var which = require('which');
 
 gulp.task('test.check-format', function() {
@@ -37,7 +38,10 @@ var tsProject =
 gulp.task('compile', function() {
   hasError = false;
   var tsResult =
-      gulp.src(['lib/**/*.ts', 'typings/**/*.d.ts', 'node_modules/typescript/lib/typescript.d.ts'])
+      gulp.src([
+            'lib/**/*.ts', 'typings/**/*.d.ts', 'node_modules/typescript/lib/typescript.d.ts',
+            'node_modules/dart-style/dart-style.d.ts'
+          ])
           .pipe(sourcemaps.init())
           .pipe(ts(tsProject))
           .on('error', onError);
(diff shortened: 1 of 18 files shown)


gulpfile.js
```diff
@@ -37,7 +38,10 @@ var tsProject =
 gulp.task('compile', function() {
   hasError = false;
   var tsResult =
-      gulp.src(['lib/**/*.ts', 'typings/**/*.d.ts', 'node_modules/typescript/lib/typescript.d.ts'])
+      gulp.src([
+            'lib/**/*.ts', 'typings/**/*.d.ts', 'node_modules/typescript/lib/typescript.d.ts',
+            'node_modules/dart-style/dart-style.d.ts'
```
mprobst: this should not be necessary, TypeScript should find this just from the `require('dart-style');`.

lib/formatter.ts
```diff
@@ -0,0 +1,19 @@
+import style = require('dart-style');
+
+interface FormatResult {
```
mprobst: Shouldn't this interface live in the dart-style .d.ts file?

lib/formatter.ts
```diff
@@ -0,0 +1,19 @@
...
+  code?: string;
+  error?: string;
+}
+
+/**
+ * Execute the Dart Formatter https://github.com/dart-lang/dart_style on either
+ * a single statement or an entire Dart library.
+ * Errors are logged if code cannot be properly formatted.
```
mprobst: Please fail, explosively? Silent failure is a total pain to debug later on.

lib/formatter.ts
jacob314: removed

lib/formatter.ts
jacob314: Moved this logic to the Transformer class and had output just return the unformatted string so I could use the existing logic to fail explosively when there was an error.

jacob314:
ptal

lib/main.ts
```diff
@@ -14,6 +14,7 @@ import StatementTranspiler from './statement';
 import TypeTranspiler from './type';
 import LiteralTranspiler from './literal';
 import {FacadeConverter} from './facade_converter';
+import * as style from 'dart-style';
```
mprobst: nit: call it `dartStyle` or so?

mprobst:
Looks pretty good. Could you check out my other comment above?

Also: in this project's `package.json`, you should now have a dependency on `dart-style` - did you run `npm install dart-style --save`?

gulpfile.js
jacob314: Fixed the module

mprobst:
Merged as c43031f.

PR #328, opened 2016-01-27
chore(facade): allow es6 methods to be defined in another location.

alexeagle:
Context: https://github.com/angular/angular/issues/5242
We don't want usage of arbitrary es6 APIs to type-check.

I'll start moving over non-es6-collections methods first, but need a ts2dart release before the angular changes.


vikerman:
LGTM

PR #330, opened 2016-02-01
feat(): Support for-of statements.

hansl:
The result is a for-in statement, which has the same semantics in Dart.


vikerman:
Merged into master.

PR #332, opened 2016-02-01
chore: Pin version of unittest Dart package used in e2e tests.

vikerman:


vikerman:
There is some problem with the versions of test/unittest that Travis is able to install. I am abandoning this PR and seeing if we can just move to "test" package.

PR #333, opened 2016-02-01
fix(test): Move e2e tests to Dart 'test' from deprecated 'unittest' package

vikerman:
Issue #331


vikerman:
Travis finally passes. This should be better than trying to control version of deprecated unittest package.

alexeagle:
awesome thank you for fixing.

PR #334, opened 2016-02-03
feat(): Add support for void generic argument, which translates to dynamic

hansl:
This allows us to use stuff like:

``` typescript
var x: Promise<void> = new Promise((resolve) => resolve());
```

Which before was harder because we couldn't use `void` as a generic type.

### DIFF
--- a/lib/base.ts
+++ b/lib/base.ts
@@ -75,6 +75,12 @@ export class TranspilerBase {
 
   maybeVisitTypeArguments(n: {typeArguments?: ts.NodeArray<ts.TypeNode>}) {
     if (n.typeArguments) {
+      // If it's a single type argument `<void>`, ignore it and emit nothing.
+      // This is particularly useful for `Promise<void>`, see
+      // https://github.com/dart-lang/sdk/issues/2231#issuecomment-108313639
+      if (n.typeArguments.length == 1 && n.typeArguments[0].kind == ts.SyntaxKind.VoidKeyword) {
+        return;
+      }
       this.emit('<');
       this.visitList(n.typeArguments);
       this.emit('>');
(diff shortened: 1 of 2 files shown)


hansl:
@vikerman @mprobst  I'll be happy to discuss this if you guys have comments/objections/questions.

lib/base.ts
```diff
@@ -76,7 +78,13 @@ export class TranspilerBase {
   maybeVisitTypeArguments(n: {typeArguments?: ts.NodeArray<ts.TypeNode>}) {
     if (n.typeArguments) {
       this.emit('<');
-      this.visitList(n.typeArguments);
+      this.visitList(n.typeArguments, ',', (node) => {
```
mprobst: As discussed offline, I think it'd be better to make this code less generic, and only specifically handle the case of a single type argument that's `void` by just leaving out the type argument.

This code is arguable better by being more generic, but in this case we should be restrictive - we shouldn't support patterns that are going to cause odd APIs for Dart users, when we really only need to support `Promise<void>` and `EventEmitter<void>`.

hansl: Done.

lib/base.ts
```diff
@@ -75,6 +75,10 @@ export class TranspilerBase {
 
   maybeVisitTypeArguments(n: {typeArguments?: ts.NodeArray<ts.TypeNode>}) {
     if (n.typeArguments) {
+      // If it's a single type argument `<void>`, ignore it and emit nothing.
```
mprobst: Nit: could you mention that this is for e.g. `Promise<void>`, and link to https://github.com/dart-lang/sdk/issues/2231#issuecomment-108313639 ?

mprobst:
One nit, LGTM otherwise.

lib/base.ts
hansl: Done. Will submit.

PR #336, opened 2016-02-05
fix(): allow facades to be in node_modules.

hansl:
When using node modules to import angular2 and use their facades, the filename isn't the one we expect. Fix that.


hansl:
@vikerman  Please take a look ASAP. This is blocking material.

hansl:
This has been merged but GitHub got confused.

PR #339, opened 2016-02-08
fix(): remove `(../)*node_modules` from every file name, and use FC for identifiers on new expressions.

hansl:


lib/facade_converter.ts
```diff
@@ -176,6 +173,7 @@ export class FacadeConverter extends base.TranspilerBase {
     var fileName = decl.getSourceFile().fileName;
     fileName = this.getRelativeFileName(fileName);
     fileName = fileName.replace(/(\.d)?\.ts$/, '');
+    fileName = fileName.replace(FACADE_NODE_MODULES_PREFIX, '');
```
alexeagle: makes more sense in this location where we were already patching up the file path

lib/facade_converter.ts
```diff
@@ -255,11 +253,7 @@ export class FacadeConverter extends base.TranspilerBase {
     'lib.es6': this.stdlibTypeReplacements,
     'angular2/typings/es6-promise/es6-promise': {'Promise': 'Future'},
     'angular2/typings/es6-shim/es6-shim': {'Promise': 'Future'},
-    '../../node_modules/rxjs/Observable': {'Observable': 'Stream'},
-    // TODO(martinprobst): It turns out the angular2 build is too eccentric to reproduce in our test
-    // suite. The ../../ path above is what happens in Angular2, the path below is what our test
-    // suite spits out.
-    'node_modules/rxjs/Observable': {'Observable': 'Stream'},
+    'rxjs/Observable': {'Observable': 'Stream'},
```
alexeagle: nice cleanup

lib/facade_converter.ts
hansl: Agreed. My last PR was a bit too targeted. I blame my insufficient knowledge of the tool; I learnt a great lot today ><

test/facade_converter_test.ts
```diff
@@ -76,11 +76,12 @@ function expectErroneousWithType(str: string) {
 describe('type based translation', () => {
   describe('Dart type substitution', () => {
     it('finds registered substitutions', () => {
-      expectWithTypes('import {Observable} from "angular2/src/facade/async"; var p: Promise<Date>;')
+      expectWithTypes(
+          'import {Observable} from "angular2/src/facade/async"; var p: Promise<Date> = new Promise<DateTime>();')
```
alexeagle: it is confusing to import Observable and not use it. That's my fault in the last change here. Can you fix that? :rose:
Just split into two expectations...

hansl: Done.

alexeagle:
still LGTM

PR #340, opened 2016-02-09
Implement an inlined facade for Promises

hansl:
After this we don't need a Promise facade, unless for `all`, `catch` and `race`. I need a handler for `then(a, b)` => `then(a).catch(b)`.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -85,10 +85,11 @@ gulp.task('test.e2e', ['test.compile'], function(done) {
   if (fs.existsSync(dir)) fsx.removeSync(dir);
   fs.mkdirSync(dir);
   fsx.copySync(__dirname + '/test/e2e', dir);
+  fsx.copySync(__dirname + '/typings', dir + '/typings');
 
   // run node with a shell so we can wildcard all the .ts files
   var cmd = 'node ../lib/main.js --translateBuiltins --basePath=. --destination=. ' +
-      '*.ts angular2/src/facade/lang.d.ts';
+      '*.ts angular2/src/facade/lang.d.ts typings/es6-promise/es6-promise.d.ts';
   // Paths must be relative to our source root, so run with cwd == dir.
   spawn('sh', ['-c', cmd], {stdio: 'inherit', cwd: dir}).on('close', function(code, signal) {
     if (code > 0) {
--- a/lib/base.ts
+++ b/lib/base.ts
@@ -14,6 +14,7 @@ export function ident(n: ts.Node): string {
 }
 
 export class TranspilerBase {
+  private id_: number = 0;
   constructor(private transpiler: Transpiler) {}
 
   visit(n: ts.Node) { this.transpiler.visit(n); }
@@ -33,6 +34,18 @@ export class TranspilerBase {
     for (var i = 0; i < nodes.length; i++) {
       this.visit(nodes[i]);
       if (i < nodes.length - 1) this.emit(separator);
+    }
+  }
+
+  uniqueId(name: string): string {
+    const id = this.id_++;
+    return `_${name}\$\$ts2dart\$${id}`;
+  }
+
+  assert(c: ts.Node, condition: boolean, reason: string): void {
+    if (!condition) {
+      this.reportError(c, reason);
+      throw new Error(reason);
     }
   }
 
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -12,6 +12,17 @@ const FACADE_DEBUG = false;
 
 const FACADE_NODE_MODULES_PREFIX = /^(\.\.\/)*node_modules\//;
 
+function merge(...args: {[key: string]: any}[]): {[key: string]: any} {
+  let returnObject: {[key: string]: any} = {};
+  for (let arg of args) {
+    for (let key of Object.getOwnPropertyNames(arg)) {
+      returnObject[key] = arg[key];
+    }
+  }
+  return returnObject;
+}
+
+
 export class FacadeConverter extends base.TranspilerBase {
   private tc: ts.TypeChecker;
   private candidateProperties: {[propertyName: string]: boolean} = {};
@@ -37,42 +48,11 @@ export class FacadeConverter extends base.TranspilerBase {
 
   maybeHandleCall(c: ts.CallExpression): boolean {
     if (!this.tc) return false;
-
-    var symbol: ts.Symbol;
-    var context: ts.Expression;
-    var ident: string;
-
-    if (c.expression.kind === ts.SyntaxKind.Identifier) {
-      // Function call.
-      ident = base.ident(c.expression);
-      if (!this.candidateProperties.hasOwnProperty(ident)) return false;
-      symbol = this.tc.getSymbolAtLocation(c.expression);
-      if (FACADE_DEBUG) console.log('s:', symbol);
-
-      if (!symbol) {
-        this.reportMissingType(c, ident);
-        return false;
-      }
-
-      context = null;
-    } else if (c.expression.kind === ts.SyntaxKind.PropertyAccessExpression) {
-      // Method call.
-      var pa = <ts.PropertyAccessExpression>c.expression;
-      ident = base.ident(pa.name);
-      if (!this.candidateProperties.hasOwnProperty(ident)) return false;
-
-      symbol = this.tc.getSymbolAtLocation(pa);
-      if (FACADE_DEBUG) console.log('s:', symbol);
-
-      // Error will be reported by PropertyAccess handling below.
-      if (!symbol) return false;
-
-      context = pa.expression;
-    } else {
-      // Not a call we recognize.
+    let {context, symbol} = this.getCallInformation(c);
+    if (!symbol) {
+      // getCallInformation returns a symbol if we understand this call.
       return false;
     }
-
     var handler = this.getHandler(c, symbol, this.callHandlers);
     return handler && !handler(c, context);
   }
@@ -128,6 +108,7 @@ export class FacadeConverter extends base.TranspilerBase {
       this.visit(typeName);
       return;
     }
+    var identifier = <ts.Identifier>typeName;
     var ident = base.ident(typeName);
     if (this.candidateTypes.hasOwnProperty(ident) && this.tc) {
       var symbol = this.tc.getSymbolAtLocation(typeName);
@@ -145,6 +126,59 @@ export class FacadeConverter extends base.TranspilerBase {
       }
     }
     this.emit(ident);
+  }
+
+  shouldEmitNew(c: ts.CallExpression): boolean {
+    if (!this.tc) return true;
+
+    let {context, symbol} = this.getCallInformation(c);
+    if (!symbol) {
+      // getCallInformation returns a symbol if we understand this call.
+      return true;
+    }
+
+    var loc = this.getFileAndName(c, symbol);
+    if (!loc) return true;
+    var {fileName, qname} = loc;
+    var fileSubs = this.callHandlerReplaceNew[fileName];
+    if (!fileSubs) return true;
+    return !fileSubs[qname];
+  }
+
+  private getCallInformation(c: ts.CallExpression): {context?: ts.Expression, symbol?: ts.Symbol} {
+    var symbol: ts.Symbol;
+    var context: ts.Expression;
+    var ident: string;
+    var expr = c.expression;
+
+    if (expr.kind === ts.SyntaxKind.Identifier) {
+      // Function call.
+      ident = base.ident(expr);
+      if (!this.candidateProperties.hasOwnProperty(ident)) return {};
+      symbol = this.tc.getSymbolAtLocation(expr);
+      if (FACADE_DEBUG) console.log('s:', symbol);
+
+      if (!symbol) {
+        this.reportMissingType(c, ident);
+        return {};
+      }
+
+      context = null;
+    } else if (expr.kind === ts.SyntaxKind.PropertyAccessExpression) {
+      // Method call.
+      var pa = <ts.PropertyAccessExpression>expr;
+      ident = base.ident(pa.name);
+      if (!this.candidateProperties.hasOwnProperty(ident)) return {};
+
+      symbol = this.tc.getSymbolAtLocation(pa);
+      if (FACADE_DEBUG) console.log('s:', symbol);
+
+      // Error will be reported by PropertyAccess handling below.
+      if (!symbol) return {};
+
+      context = pa.expression;
+    }
+    return {context, symbol};
   }
 
   private getHandler<T>(n: ts.Node, symbol: ts.Symbol, m: ts.Map<ts.Map<T>>): T {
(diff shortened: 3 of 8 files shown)


lib/base.ts
```diff
@@ -36,6 +36,12 @@ export class TranspilerBase {
     }
   }
 
+  assert(condition: boolean, reason?: string): void {
```
mprobst: please mark the reason as required - assertions without reasons suck.

lib/facade_converter.ts
```diff
@@ -147,6 +147,48 @@ export class FacadeConverter extends base.TranspilerBase {
...
+      symbol = this.tc.getSymbolAtLocation(c.expression);
+      if (FACADE_DEBUG) console.log('s:', symbol);
+
+      if (!symbol) {
+        this.reportMissingType(c, ident);
+        return true;
+      }
+    } else if (c.expression.kind === ts.SyntaxKind.PropertyAccessExpression) {
+      // Method call.
+      var pa = <ts.PropertyAccessExpression>c.expression;
```
mprobst: Can you de-duplicate this code some more? I think if you create a variable for `ident` and `symbolLocation` or so, you can share most of the code.

lib/facade_converter.ts
```diff
@@ -354,6 +396,48 @@ export class FacadeConverter extends base.TranspilerBase {
...
+      this.emit('Future.error(');
+      this.visitList(c.arguments);  // Should only be 1.
+      this.emit(')');
+    },
+    'Promise': (c: ts.CallExpression, context: ts.Expression) => {
+      if (c.kind != ts.SyntaxKind.NewExpression) {
+        return true;
+      }
+      this.assert(c.arguments.length == 1);
+      const callback = <ts.ArrowFunction>(c.arguments[0]);
```
mprobst: this seems like a risky cast - might be a function, as you assert below, right? Do you need an `ArrowFunction`, or some super type?

lib/facade_converter.ts
```diff
@@ -354,6 +396,48 @@ export class FacadeConverter extends base.TranspilerBase {
...
+      this.assert(
+          callback.kind == ts.SyntaxKind.ArrowFunction ||
+          callback.kind == ts.SyntaxKind.FunctionDeclaration ||
+          callback.kind == ts.SyntaxKind.FunctionExpression);
+      this.assert(callback.parameters.length > 0 && callback.parameters.length < 3);
+
+      this.emit('(() {');  // Create a new scope.
+      this.emit('Completer _completer = new Completer();');
```
mprobst: This code is too clever. You also risk generating code that's substantially larger than the original source.

I think we should use a helper method for this.

test/facade_converter_test.ts
```diff
@@ -54,7 +69,7 @@ function getSources(str: string): {[k: string]: string} {
           static get(m: any, k: string): number { return m[k]; }
         }
         var global: any;
-        export var Promise = global.Promise;`,
+        export var Promise = 1;`,
```
mprobst: huh?

test/facade_converter_test.ts
```diff
@@ -36,8 +36,23 @@ function getSources(str: string): {[k: string]: string} {
     'angular2/src/core/di/forward_ref.d.ts': `
         export declare function forwardRef<T>(x: T): T;`,
     'angular2/typings/es6-promise/es6-promise.d.ts': `
-        declare class Promise<R> {}
-        declare module Promise {}`,
+      interface Thenable<R> {
+        then<U>(onFulfilled?: (value: R) => U | Thenable<U>,  onRejected?: (error: any) => U | Thenable<U>): Thenable<U>;
+      }
+      declare class Promise<R> implements Thenable<R> {
```
mprobst: Do you need all of these methods?

test/facade_converter_test.ts
```diff
@@ -36,8 +36,23 @@ function getSources(str: string): {[k: string]: string} {
...
+        then<U>(onFulfilled?: (value: R) => U | Thenable<U>,  onRejected?: (error: any) => U | Thenable<U>): Promise<U>;
+        catch<U>(onRejected?: (error: any) => U | Thenable<U>): Promise<U>;
+        static resolve<R>(value?: R | Thenable<R>): Promise<R>;
+        static reject(error: any): Promise<any>;
+        static all<R>(promises: (R | Thenable<R>)[]): Promise<R[]>;
+      }
+      declare var Promise: {
```
mprobst: Do you really need the class/var dualism here? You could also just declare a ctor in the class above, I think.

test/facade_converter_test.ts
```diff
@@ -36,8 +36,23 @@ function getSources(str: string): {[k: string]: string} {
...
+        new(callback: (resolve : (value?: R | Thenable<R>) => void, reject: (error?: any) => void) => void);
+        prototype: Promise<any>;
+      };
+      declare module Promise {
+        function race<R>(promises: (R | Thenable<R>)[]): Promise<R>;
```
mprobst: I think you could just make this a static function above.

test/facade_converter_test.ts
```diff
@@ -36,8 +36,23 @@ function getSources(str: string): {[k: string]: string} {
     'angular2/src/core/di/forward_ref.d.ts': `
         export declare function forwardRef<T>(x: T): T;`,
     'angular2/typings/es6-promise/es6-promise.d.ts': `
-        declare class Promise<R> {}
-        declare module Promise {}`,
+      interface Thenable<R> {
+        then<U>(onFulfilled?: (value: R) => U | Thenable<U>,  onRejected?: (error: any) => U | Thenable<U>): Thenable<U>;
```
mprobst: Re all the comments below, a perfectly fine answer is "I'm just copying whatever is in lib.d.ts". But then please add a comment.

test/facade_converter_test.ts
```diff
@@ -237,6 +248,52 @@ var y = x.length;`);
 }`);
   });
 
+  describe('promises', () => {
+    it('translates into Futures', () => {
+      expectWithTypes('let x: Promise = Promise.resolve(1);').to.equal(`import "dart:async";
+
+Future x = Future.value(1);`);
```
mprobst: Nice!

test/facade_converter_test.ts
```diff
@@ -237,6 +248,52 @@ var y = x.length;`);
...
+      expectWithTypes('let x: Promise = Promise.reject(1);').to.equal(`import "dart:async";
+
+Future x = Future.error(1);`);
+      expectWithTypes('let x: Promise = new Promise((resolve) => {resolve(1);});')
+          .to.equal(`import "dart:async";
+
+Future x = (() {
```
mprobst: as discussed, I think we should require a helper function for this.

I presume there is nothing in `dart:async` already that would have the same semantics?

mprobst:
Nice. Needs a bit more work/discussion, but good to see that stuff go away!

lib/facade_converter.ts
hansl: Actually I need a `FunctionLikeDeclaration`. I clarified the code, but there's no common super class.

lib/facade_converter.ts
hansl: I agree with you, but the problem is deciding where that code live. I don't mind a PromiseWrapper but we shouldn't include it anywhere. @alexeagle @mprobst What's your opinion on this? Should I create a library that's included in the pub dependencies (a kind of `ts2dart/stdlib`) or should I add a helper function in this file scope?

test/facade_converter_test.ts
hansl: Our code is now clever enough that we detect aliases.

test/facade_converter_test.ts
hansl: It's just the official interface of Promises. I still need to add `then` support. I removed the Thenable interface though.

test/facade_converter_test.ts
hansl: Didn't work before. Let me look around.

test/facade_converter_test.ts
hansl: This should be static, yes. But for now it's unimplemented.

test/facade_converter_test.ts
hansl: Haha. Yes, that's what I'm basically doing. But I removed `Thenable` as it's unnecessary.

test/facade_converter_test.ts
hansl: The Completer object is the closest to something that works. This trick is a lot more code but keeps the interface intact. I can move it somewhere else but that would require the client to have a dependency to a library. It would also assume we need to emit code in the global scope... which I think we don't support right now.

hansl:
I reduced the generated code quite a bit. It's now closer to what someone would write if they were to implement the code by hand (assuming they couldn't change the function content). I think this is good enough for now, with a potential optimization of pushing the code in a function helper.

lib/base.ts
hansl: I also added a context so we would report the error with line:column and such.

lib/facade_converter.ts
hansl: Done.

hansl:
@alexeagle @mprobst PTAL.

lib/base.ts
```diff
@@ -36,6 +36,13 @@ export class TranspilerBase {
     }
   }
 
+  assert(c: ts.Node, condition: boolean, reason: string): void {
+    if (!condition) {
+      this.reportError(c, reason);
+      throw new Error(reason);
```
mprobst: actually, isn't this just `reportError` now? Maybe just `reportError(...); return;` instead of assert?

hansl: Assertions should stop the execution flow. In the case of Promises (where it's first used), assertion means that we already know the code generated will not work.

test/facade_converter_test.ts
mprobst: Still don't get it :-)

So the problem is that we find a global variable called Promise, but before it was an alias, and we resolved through the alias? Maybe change it to something like `{definedHere: true}` and add a comment?

lib/facade_converter.ts
```diff
@@ -147,6 +118,58 @@ export class FacadeConverter extends base.TranspilerBase {
...
+
+  private getCallInformation(c: ts.CallExpression): {context?: ts.Expression, symbol?: ts.Symbol} {
+    var symbol: ts.Symbol;
+    var context: ts.Expression;
+    var ident: string;
+
+    if (c.expression.kind === ts.SyntaxKind.Identifier) {
+      // Function call.
+      ident = base.ident(c.expression);
+      if (!this.candidateProperties.hasOwnProperty(ident)) return {};
```
mprobst: This block still duplicates a lot of code from the `if` block immediately following it. Can't you assign into `ident` and `symbol` in the if statements, and then execute the same code as below, potentially with a check when to return an error vs an empty object?

lib/facade_converter.ts
```diff
@@ -354,6 +377,73 @@ export class FacadeConverter extends base.TranspilerBase {
...
+      if (c.kind != ts.SyntaxKind.NewExpression) {
+        return true;
+      }
+      this.assert(c, c.arguments.length == 1, 'Promise construction must take 2 arguments.');
+      this.assert(
+          c, c.arguments[0].kind == ts.SyntaxKind.ArrowFunction ||
+              c.arguments[0].kind == ts.SyntaxKind.FunctionExpression,
+          'Promise argument must be a function expression (or arrow function).');
+      let callback: ts.FunctionLikeDeclaration;
+      if (c.arguments[0].kind == ts.SyntaxKind.ArrowFunction)
```
mprobst: for multi-line if statements, always use braces. same below.

lib/facade_converter.ts
```diff
@@ -354,6 +377,73 @@ export class FacadeConverter extends base.TranspilerBase {
...
+      this.emit(')');
+    },
+    'Promise.catch': (c: ts.CallExpression, context: ts.Expression) => {
+      this.visit(context);
+      this.emit('.catchError(');
+      this.visitList(c.arguments);
+      this.emit(')');
+    },
+    'Promise': (c: ts.CallExpression, context: ts.Expression) => {
+      if (c.kind != ts.SyntaxKind.NewExpression) {
```
mprobst: for single line if statements, it's ok to keep them in line:
```js
if (foo) return false;
```

lib/facade_converter.ts
```diff
@@ -354,6 +377,73 @@ export class FacadeConverter extends base.TranspilerBase {
...
+        return true;
```
mprobst: nit: can you add a comment explaining why it's ok to return here?

lib/facade_converter.ts
hansl: Your clang format made this...

lib/facade_converter.ts
mprobst: I think having a Dart library like `ts2dart/support` (or `stdlib` or `facades` or whatever, all works for me) sounds like a good idea. Though please ask somebody with more Dart experience than me.

lib/facade_converter.ts
hansl: These are two subtly different cases. In the case of an Identifier, we just need the symbol of its expression, while for a PropertyAccessExpression we need the symbol of the expression preceding it. Also if its neither the code shouldn't do anything. They're not so easily de-duplication-able.

lib/base.ts
mprobst: Yeah assertions should, but in `ts2dart` it has turned out that it's super useful to continue program execution to find *all* the errors in one go (or as many as possible), so that users don't have to laboriously fix errors one by one, and then re-run the (rather slow) tool.

In particular things like Promise ctor parameter count, or callback parameter count are things that might end up being incorrect in user code for some reason or another. We should not assert there, we should reportError as much as possible.

mprobst:
I agree that with the reduced code generation, this is ok without a library/helper func. We should keep that idea in mind though, that would be the better solution to Angular2's façade file.

lib/facade_converter.ts
```diff
@@ -354,6 +377,73 @@ export class FacadeConverter extends base.TranspilerBase {
...
+          c, callback.parameters.length > 0 && callback.parameters.length < 3,
+          'Promise executor must take 1 or 2 arguments (resolve and reject).');
+
+      this.assert(
+          c, callback.parameters[0].name.kind == ts.SyntaxKind.Identifier,
+          'First argument of the Promise executor is not a straight parameter.');
+      let resolveParameterIdent = <ts.Identifier>(callback.parameters[0].name);
+      this.assert(
+          c, resolveParameterIdent.text != '_completer',
+          'Promise executor arguments cannot have "_completer" as a name.');
```
mprobst: actually the real fix is to ensure there's no other `_completer` in scope. Maybe make the variable a bit more unique, `_completer$ts2dart` or so?

lib/facade_converter.ts
```diff
@@ -354,6 +377,73 @@ export class FacadeConverter extends base.TranspilerBase {
...
+
+      this.emit('(() {');  // Create a new scope.
```
mprobst: AFAIK dart2js does not inline IIFEs.

Can we do away with the extra scope, assuming the completer var name is unique enough?

mprobst:
Looking really good already.

Given that we generate somewhat complicated Dart code, could you add an integration test in the e2e folder? That would ensure the generated code actually works...

lib/facade_converter.ts
mprobst: So keep the assignment into `expression` in the `if` branch for each node, and move the common code out of the `if` blocks?

lib/facade_converter.ts
alexeagle: good point, this may be performance-sensitive generated code

alexeagle:
I'm not going to review in any detail since @mprobst is giving an excellent thorough review already ;)

hansl:
I've done the e2e test.

lib/facade_converter.ts
hansl: Did one better, and trying to get rid of the `(() {})()` function call.

lib/facade_converter.ts
hansl: Actually I cannot get rid of that scope for various reason:

1. The point where I'm at in the emit stream is `Future someVarName = >>>here<<<`. Which means that I can't rewrite the declaration.
2. Having access to the `Future` variable name would be really hard as well, because multiple places in the ancestry could be defining it. It could be a key on an object... :/ Or it could be entirely unnamed.
3. I've re-added the inner function scope because the executor might return, and in this case that would break the generated code :/

If you find solutions for some of these I can get it in a better shape. Unless we actually do AST => AST transform at some point this will be a really hard problem.

mprobst: Yeah, let's go with this for the time being.

lib/base.ts
```diff
@@ -14,6 +14,7 @@ export function ident(n: ts.Node): string {
 }
 
 export class TranspilerBase {
+  private id_: number = 0;
```
mprobst: nit: we don't use `_` for private fields here.

test/facade_converter_test.ts
```diff
@@ -237,6 +243,99 @@ var y = x.length;`);
...
+    resolve(1);
+  })();
+  return _completer$$ts2dart$0.future;
+})();`);
+      expectWithTypes('let x: Promise = new Promise((myParam1, myParam2) => {myParam1(1);});')
+          .to.equal(`import "dart:async";
+
+Future x = (() {
+  Completer _completer$$ts2dart$0 = new Completer();
+  var myParam1 = _completer$$ts2dart$0.complete;
```
alexeagle: Kind of sad for angular dart users that the library sources they browse are now more obviously generated code

test/facade_converter_test.ts
```diff
@@ -35,9 +35,19 @@ function getSources(str: string): {[k: string]: string} {
     'angular2/typings/es6-shim/es6-shim': es6RuntimeDeclarations,
     'angular2/src/core/di/forward_ref.d.ts': `
         export declare function forwardRef<T>(x: T): T;`,
-    'angular2/typings/es6-promise/es6-promise.d.ts': `
-        declare class Promise<R> {}
-        declare module Promise {}`,
+    'typings/es6-promise/es6-promise.d.ts': `
```
alexeagle: doesn't have to be in this change, but it's clear we should just fetch typings in the postinstall for this project, and pull the content from it, instead of copying it in here

test/facade_converter_test.ts
alexeagle: Can you just revert the line instead?

gulpfile.js
```diff
@@ -88,7 +88,7 @@ gulp.task('test.e2e', ['test.compile'], function(done) {
 
   // run node with a shell so we can wildcard all the .ts files
   var cmd = 'node ../lib/main.js --translateBuiltins --basePath=. --destination=. ' +
-      '*.ts angular2/src/facade/lang.d.ts';
+      '*.ts angular2/src/facade/lang.d.ts typings/es6-promise/es6-promise.d.ts';
```
alexeagle: need to install this file in a postinstall

test/facade_converter_test.ts
```diff
@@ -237,6 +243,99 @@ var y = x.length;`);
...
+
+Future x = (() {
+  Completer _completer$$ts2dart$0 = new Completer();
+  var resolve = _completer$$ts2dart$0.complete;
+  var reject = _completer$$ts2dart$0.completeError;
+  (() {
+    resolve(1);
+  })();
+  return _completer$$ts2dart$0.future;
+})();`);
```
alexeagle: discussed in person: this function execution doesn't happen in the hand-written angular2 promisewrapper code, so this may have a performance implication.
But Hans has an idea to remove this by passing the resolve and reject as arguments to the inner IIFE

test/facade_converter_test.ts
hansl: Replaced with `class Promise` which is what the test is checking for anyway.

test/facade_converter_test.ts
hansl: After review, without AST transform this will be impossible. The problem is that the code preceding the syntax emission point can be any value left hand side expression, so we can't change it or copy it properly.

The transform here would be simple that maybe it could be done in a pass; create the Completer before the Future declaration and change the value of the future by `completer.future`. This would remove the first function call.

test/facade_converter_test.ts
hansl: Ack.

gulpfile.js
hansl: Added a setting on travis. We can't use a postinstall as we publish this package.

alexeagle: wait, we should not need to add anything, there is already a tsd.json in the project, just need to add es6-promise to it.

.travis.yml
```diff
@@ -4,6 +4,7 @@ node_js:
   - "stable"
 before_script:
   - npm install
+  - npm run typings
```
alexeagle: should not need this, see below

gulpfile.js
hansl: That was done a while ago though. Why is it not finding the `d.ts`?

alexeagle: because es6-promise is not there?
https://github.com/angular/ts2dart/blob/master/tsd.json

hansl: That's in master, not my PR. My PR adds it.

lib/type.ts
```diff
@@ -70,7 +70,8 @@ export default class TypeTranspiler extends base.TranspilerBase {
         break;
       case ts.SyntaxKind.Identifier:
         var ident = <ts.Identifier>node;
-        this.emit(ident.text);
+        this.fc.visitTypeName(ident);
+        // this.emit(ident.text);
```
alexeagle: remove comment?

test/facade_converter_test.ts
```diff
@@ -1,4 +1,5 @@
 /// <reference path="../typings/mocha/mocha.d.ts"/>
+import fs = require('fs');
```
alexeagle: I think this should be import * as fs from 'fs'

hansl: Done.

lib/type.ts
hansl: Done.

PR #343, opened 2016-02-17
Sanitize TypeScript doc strings to remove tags not recognized by dartdoc

### LINKED ISSUE #341: Sanitize Typescript document comments to remove tags not supported by dartdoc
The following doc annotations don't mean anything for dartdoc and need to be removed/changed

\@module \@description
\@param
\@deprecated

vikerman:
Fixes #341

### DIFF
--- a/lib/main.ts
+++ b/lib/main.ts
@@ -275,7 +275,19 @@ export class Transpiler {
   private normalizeSlashes(path: string) { return path.replace(/\\/g, '/'); }
 
   private translateComment(comment: string): string {
-    return comment.replace(/\{@link ([^\}]+)\}/g, '[$1]');
+    comment = comment.replace(/\{@link ([^\}]+)\}/g, '[$1]');
+
+    // Remove the following tags and following comments till end of line.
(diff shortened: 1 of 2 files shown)


lib/main.ts
```diff
@@ -275,7 +275,19 @@ export class Transpiler {
   private normalizeSlashes(path: string) { return path.replace(/\\/g, '/'); }
 
   private translateComment(comment: string): string {
-    return comment.replace(/\{@link ([^\}]+)\}/g, '[$1]');
+    comment = comment.replace(/\{@link ([^\}]+)\}/g, '[$1]');
+
+    // Remove the following tags and following comments till end of line.
+    comment = comment.replace(/@param.*$/gm, '');
```
alexeagle: opt: what about just chaining the .replace calls instead of reassigning to the comment variable?

return comment.replace(/1/).replace(/2/).replace...
unless having separate statements is essential for stepping with a debugger?

alexeagle:
LGTM

PR #344, opened 2016-02-18
fix(): add a lenientTypingFileCheck option to only check the end of the path

hansl:


mprobst:
This seems a bit dangerous. What's the problem that you're actually trying to solve? Maybe there's a solution that's on a slightly higher semantic level, e.g. search with or without `node_modules`?

I think the name of the feature is also quite misleading, maybe `matchFullFacadePath: false` is more speaking?

hansl:
This is bad in general, I agree with you. After talking with Alex I'm going to come up with a `typingsRoot` argument instead.

PR #345, opened 2016-02-18
fix(): allow the caller (or CLI) to specify a root directory for typings

hansl:
This allows the caller to put the typings anywhere and still be able to use us without waiting for a release of ts2dart.

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -27,12 +27,15 @@ export class FacadeConverter extends base.TranspilerBase {
   private tc: ts.TypeChecker;
   private candidateProperties: {[propertyName: string]: boolean} = {};
   private candidateTypes: {[typeName: string]: boolean} = {};
+  private typingsRootRegex: RegExp;
 
-  constructor(transpiler: Transpiler) {
+  constructor(transpiler: Transpiler, typingsRoot: string = '') {
     super(transpiler);
     this.extractPropertyNames(this.callHandlers, this.candidateProperties);
     this.extractPropertyNames(this.propertyHandlers, this.candidateProperties);
     this.extractPropertyNames(this.TS_TO_DART_TYPENAMES, this.candidateTypes);
+
+    this.typingsRootRegex = new RegExp('^' + typingsRoot.replace('.', '\\.'));
   }
 
   private extractPropertyNames(m: ts.Map<ts.Map<any>>, candidates: {[k: string]: boolean}) {
@@ -204,10 +207,11 @@ export class FacadeConverter extends base.TranspilerBase {
       decl = symbol.declarations[0];
     }
 
-    var fileName = decl.getSourceFile().fileName;
-    fileName = this.getRelativeFileName(fileName);
-    fileName = fileName.replace(/(\.d)?\.ts$/, '');
-    fileName = fileName.replace(FACADE_NODE_MODULES_PREFIX, '');
+    const fileName = decl.getSourceFile().fileName;
+    const canonicalFileName = this.getRelativeFileName(fileName)
+                                  .replace(/(\.d)?\.ts$/, '')
+                                  .replace(FACADE_NODE_MODULES_PREFIX, '')
+                                  .replace(this.typingsRootRegex, '');
 
     var qname = this.tc.getFullyQualifiedName(symbol);
     // Some Qualified Names include their file name. Might be a bug in TypeScript,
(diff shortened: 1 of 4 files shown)


hansl:
I'm testing this with Material2 to make sure it solves the problem.

test/facade_converter_test.ts
```diff
@@ -107,6 +123,13 @@ ByteBuffer buff;`);
 
 var y = x is Promise;`);
     });
+
+    it('works with a different typings directory', () => {
+      expectWithTypingsRootTypes('var y: Promise = Promise.resolve(1);')
```
alexeagle: could it just `import {Promise} from 'somewhere'`
and add a somewhere.d.ts in the existing getSources fixture above?

lib/facade_converter.ts
```diff
@@ -209,6 +214,10 @@ export class FacadeConverter extends base.TranspilerBase {
     fileName = fileName.replace(/(\.d)?\.ts$/, '');
     fileName = fileName.replace(FACADE_NODE_MODULES_PREFIX, '');
 
+    if (this.typingsRootRegex != null) {
+      fileName = fileName.replace(this.typingsRootRegex, '');
```
alexeagle: can we make another local variable `canonicalFileName` so it's more clear what is happening

lib/facade_converter.ts
```diff
@@ -285,11 +294,14 @@ export class FacadeConverter extends base.TranspilerBase {
   private TS_TO_DART_TYPENAMES: ts.Map<ts.Map<string>> = {
     'lib': this.stdlibTypeReplacements,
     'lib.es6': this.stdlibTypeReplacements,
-    'typings/es6-promise/es6-promise': {'Promise': 'Future'},
     'angular2/typings/es6-promise/es6-promise': {'Promise': 'Future'},
```
alexeagle: if we default typingsRoot to angular2/typings then we could remove a couple of these?

alexeagle:
thanks this seems like a great direction

alexeagle:
FYI @vikerman

test/facade_converter_test.ts
hansl: The `getSources` files are compiled along the file, so they're ambient.

Do you want to falsify the test? Is this what you want? I could verify that `import {Promise} from 'another/path/to/es6-promise/es6-promise';` would not cause it to be renamed.

lib/facade_converter.ts
hansl: It might break `angular2`, but if you can test it, then I'll do it. I just prefer smaller deltas.

lib/facade_converter.ts
hansl: Done.

test/facade_converter_test.ts
alexeagle: No it's the same logic you have now, I'm just saying that if you add another Promise symbol in another place, and set the prefix to point to that place, you can keep this same assertion with less duplicate fixture. Does that make sense?

lib/main.ts
```diff
@@ -40,6 +40,10 @@ export interface TranspilerOptions {
    * Enforce conventions of public/private keyword and underscore prefix
    */
   enforceUnderscoreConventions?: boolean;
+  /**
+   * Sets the root path for all typings used by the facade converter.
```
alexeagle: technically it's just *a* root path, right?

lib/main.ts
```diff
@@ -62,7 +66,7 @@ export class Transpiler {
   private fc: FacadeConverter;
 
   constructor(private options: TranspilerOptions = {}) {
-    this.fc = new FacadeConverter(this);
+    this.fc = new FacadeConverter(this, options.typingsRoot || 'angular2/typings/');
```
alexeagle: Leave a TODO to remove after transition

test/facade_converter_test.ts
```diff
@@ -33,10 +33,10 @@ var es6RuntimeDeclarations = `
 
 function getSources(str: string): {[k: string]: string} {
   var srcs: {[k: string]: string} = {
-    'angular2/typings/es6-shim/es6-shim': es6RuntimeDeclarations,
+    'some/path/to/typings/es6-shim/es6-shim': es6RuntimeDeclarations,
```
alexeagle: yes, much simpler :)

alexeagle:
LGTM

lib/main.ts
hansl: Done.

hansl:
Merging as soon as travis is green.

PR #350, opened 2016-03-02
DDC style generic method

mprobst:


lib/main.ts
```diff
@@ -4,7 +4,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import * as ts from 'typescript';
 
-import {TranspilerBase} from './base';
+import {TranspilerBase, ident} from './base';
```
vikerman: Where is ident used?

vikerman:
LGTM

lib/main.ts
mprobst: Whoops, leftover, removed.

PR #351, opened 2016-03-02
Switch to version of dart-style that does not pollute global

jacob314:
Let me know where you would like the regression test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "test": "test"
   },
   "dependencies": {
-    "dart-style": "^0.2.6",
+    "dart-style": "^0.2.7",
     "minimist": "^1.1.1",
     "source-map": "^0.4.2",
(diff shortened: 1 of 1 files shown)


test/main_test.ts
```diff
@@ -83,3 +83,11 @@ class A {
     });
   });
 });
+
+describe('regression', () => {
```
mprobst: I think it's fine to leave out the test. It's a corner case internal bug in a dependency - this should be a test in dart-style, but not in ts2dart.

jacob314: Reverted the test file. This is really a test for the library that wraps dart2js output so it can run on NodeJS.

vikerman:
LGTM. Will merge it soon. Thanks

mprobst:
Merged in c313fbb.

PR #353, opened 2016-03-09
fix: ensure that numbers are temporarily casted to ints for bitwise operations

matsko:


lib/expression.ts
```diff
@@ -20,22 +20,83 @@ export default class ExpressionTranspiler extends base.TranspilerBase {
...
+            case ts.SyntaxKind.BarEqualsToken:
+            case ts.SyntaxKind.AmpersandEqualsToken:
+            case ts.SyntaxKind.GreaterThanGreaterThanEqualsToken:
+            case ts.SyntaxKind.LessThanLessThanEqualsToken:
+              this.visit(binExpr.left);
+              this.emit('=');
+              this.emit('(');
+              this.visit(binExpr.left);
+              this.emit('as int)');
+              this.emit(tokenStr.substring(0, tokenStr.indexOf('=')));
```
mprobst:
```
'<<='.slice(0, -1)
'<<'
```

Also, add a comment why this is happening

lib/expression.ts
```diff
@@ -20,22 +20,83 @@ export default class ExpressionTranspiler extends base.TranspilerBase {
...
+              if (binExpr.right.kind == ts.SyntaxKind.Identifier) {
+                this.emit('(');
+                this.visit(binExpr.right);
+                this.emit('as int)');
+              } else {
+                this.visit(binExpr.right);
+              }
+              break;
+
```
mprobst: nit: drop the line here

lib/expression.ts
```diff
@@ -20,22 +20,83 @@ export default class ExpressionTranspiler extends base.TranspilerBase {
...
+            case ts.SyntaxKind.AmpersandEqualsToken:
+            case ts.SyntaxKind.GreaterThanGreaterThanEqualsToken:
+            case ts.SyntaxKind.LessThanLessThanEqualsToken:
+              this.visit(binExpr.left);
+              this.emit('=');
+              this.emit('(');
+              this.visit(binExpr.left);
+              this.emit('as int)');
+              this.emit(tokenStr.substring(0, tokenStr.indexOf('=')));
+              if (binExpr.right.kind == ts.SyntaxKind.Identifier) {
```
mprobst: I think this is not quite correct (think of e.g. function calls or whatever), and not strictly needed. Just drop it, and always emit the case.

lib/expression.ts
```diff
@@ -20,22 +20,83 @@ export default class ExpressionTranspiler extends base.TranspilerBase {
           this.visit(binExpr.right);
```
mprobst: Merge all of this into one switch statement, `if` on some expression followed by `switch` on the same expression is odd.

lib/expression.ts
```diff
@@ -20,22 +20,83 @@ export default class ExpressionTranspiler extends base.TranspilerBase {
...
-          if (operatorKind === ts.SyntaxKind.InstanceOfKeyword) {
-            this.emit('is');
-            this.fc.visitTypeName(<ts.Identifier>binExpr.right);
-          } else if (operatorKind == ts.SyntaxKind.InKeyword) {
-            this.reportError(node, 'in operator is unsupported');
-          } else {
-            this.emit(ts.tokenToString(binExpr.operatorToken.kind));
-            this.visit(binExpr.right);
+          var tokenStr = ts.tokenToString(operatorKind);
+          switch (operatorKind) {
```
mprobst: Needs a comment explaining this is handling bit fiddling operators, and why you need the explicit cast for Dart.

lib/expression.ts
```diff
@@ -20,22 +20,83 @@ export default class ExpressionTranspiler extends base.TranspilerBase {
...
+              this.visit(binExpr.left);
+              this.emit('=');
+              this.emit('(');
+              this.visit(binExpr.left);
+              this.emit('as int)');
+              this.emit(tokenStr.substring(0, tokenStr.indexOf('=')));
+              if (binExpr.right.kind == ts.SyntaxKind.Identifier) {
+                this.emit('(');
+                this.visit(binExpr.right);
+                this.emit('as int)');
```
mprobst: Extract a method for these three emits

lib/expression.ts
```diff
@@ -20,22 +20,83 @@ export default class ExpressionTranspiler extends base.TranspilerBase {
...
-        this.emit(ts.tokenToString(prefixUnary.operator));
-        this.visit(prefixUnary.operand);
+        var operator = ts.tokenToString(prefixUnary.operator);
+        this.emit(operator);
+
+        if (prefixUnary.operator == ts.SyntaxKind.TildeToken &&
+            prefixUnary.operand.kind == ts.SyntaxKind.Identifier) {
+          this.emit('(');
+          this.visit(prefixUnary.operand);
+          this.emit('as int)');
```
mprobst: use the method

package.json
```diff
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "dart-style": "^0.2.6",
-    "minimist": "^1.1.1",
+    "minimist": "^1.2.0",
```
mprobst: please revert

test/expression_test.ts
```diff
@@ -92,4 +92,90 @@ describe('expressions', () => {
     expectTranslate('foo.bar;').to.equal('foo.bar;');
     expectTranslate('foo[bar];').to.equal('foo[bar];');
   });
+  it('should cast numeric values to int when bitwise operators are used for a LHS exp RHS numeric expression',
+     () => {
+       function makeExpr(operator: string) {
+         return 'var x: number = 1; var y: number = 2; var z: number = x ' + operator + ' y;';
```
mprobst: use a template string

test/expression_test.ts
```diff
@@ -92,4 +92,90 @@ describe('expressions', () => {
...
+       }
+
+       expectTranslate(makeExpr('|'))
+           .to.equal("num x = 1;\nnum y = 2;\nnum z = (x as int) | (y as int);");
```
mprobst: consider just having the `num z = (x as int) | (y as int)` part, but dropping the declarations before, I think they are not needed.

test/expression_test.ts
```diff
@@ -92,4 +92,90 @@ describe('expressions', () => {
     expectTranslate('foo.bar;').to.equal('foo.bar;');
     expectTranslate('foo[bar];').to.equal('foo[bar];');
   });
+  it('should cast numeric values to int when bitwise operators are used for a LHS exp RHS numeric expression',
+     () => {
+       function makeExpr(operator: string) {
```
mprobst: consider doing a table as above

test/expression_test.ts
```diff
@@ -92,4 +92,90 @@ describe('expressions', () => {
...
+           .to.equal("num x = 1;\nnum y = 2;\nnum z = (x as int) >> (y as int);");
+     });
+  it('should only cast identifier values when consumed in a bitwise expression', () => {
+    function makeExpr(operator: string) {
+      return 'function f() { var delims: number = 100; delims = delims ' + operator + ' 128 ' +
+          operator + ' 256; }';
+    }
+
+    expectTranslate(makeExpr('|'))
+        .to.equal("f() {\n  num delims = 100;\n  delims = (delims as int) | 128 | 256;\n}");
```
mprobst: drop this after changing original

test/expression_test.ts
```diff
@@ -92,4 +92,90 @@ describe('expressions', () => {
...
+       function makeExpr(operator: string) {
+         return 'function f() { var delims: number = 100; var tweaks: number = 128; g(delims ' +
+             operator + ' tweaks ' + operator + ' 256); }';
+       }
+
+       expectTranslate(makeExpr('|'))
+           .to.equal(
+               "f() {\n  num delims = 100;\n  num tweaks = 128;\n  g((delims as int) | (tweaks as int) | 256);\n}");
+
+       expectTranslate(makeExpr('&'))
```
mprobst: Consider inlining the expression equals:

```
num x = 1; // probably don't need this
num y = (x |= 2);
```

lib/expression.ts
```diff
@@ -11,31 +17,71 @@ export default class ExpressionTranspiler extends base.TranspilerBase {
...
+          case ts.SyntaxKind.BarToken:
+          case ts.SyntaxKind.AmpersandToken:
+          case ts.SyntaxKind.GreaterThanGreaterThanToken:
+          case ts.SyntaxKind.LessThanLessThanToken:
+          case ts.SyntaxKind.CaretEqualsToken:
+          case ts.SyntaxKind.BarEqualsToken:
+          case ts.SyntaxKind.AmpersandEqualsToken:
+          case ts.SyntaxKind.GreaterThanGreaterThanEqualsToken:
+          case ts.SyntaxKind.LessThanLessThanEqualsToken:
+            // we special case this for Dart since dart2ts will
```
mprobst: nit: no `we` in comments, always use third party voice.

mprobst: dart2ts?

Maybe formulate as `the bitwise operators are only defined on int, so the number types ts2dart deals with have to be converted to int explicitly to match JavaScript's semantics in Dart`?

lib/expression.ts
```diff
@@ -11,31 +17,71 @@ export default class ExpressionTranspiler extends base.TranspilerBase {
...
+            // convert `number` to `num` and if there are any numbers
+            // that are evaluated with bitwise operations then the `num`
+            // number type will be handled properly. The code below will
+            // explicitly cast the associated values into ints temporarily.
+
+            if (tokenStr[tokenStr.length - 1] == "=") {
+              // assignment case (ident [op]=)
```
mprobst: add a comment on what this is doing? `For assignments, strip the trailing = so that...`

lib/expression.ts
```diff
@@ -11,31 +17,71 @@ export default class ExpressionTranspiler extends base.TranspilerBase {
...
         }
         break;
       case ts.SyntaxKind.PrefixUnaryExpression:
         var prefixUnary = <ts.PrefixUnaryExpression>node;
-        this.emit(ts.tokenToString(prefixUnary.operator));
-        this.visit(prefixUnary.operand);
+        var operator = ts.tokenToString(prefixUnary.operator);
+        this.emit(operator);
+
+        if (prefixUnary.operator == ts.SyntaxKind.TildeToken) {
```
mprobst: nit: `===`

test/expression_test.ts
```diff
@@ -92,4 +92,59 @@ describe('expressions', () => {
     expectTranslate('foo.bar;').to.equal('foo.bar;');
     expectTranslate('foo[bar];').to.equal('foo[bar];');
   });
+  it('should cast numeric values to int when bitwise operators are used for a LHS exp RHS numeric expression',
+     () => {
+       expectTranslate(`var z: number = x | y`).to.equal("num z = (x as int) | (y as int);");
```
mprobst: this and the stuff below might also look better as a table, but your call.

test/expression_test.ts
```diff
@@ -92,4 +92,59 @@ describe('expressions', () => {
...
+       expectTranslate('var y: number = (y ^= x)')
+           .to.equal('num y = (y = (y as int) ^ (x as int));');
+
+       expectTranslate('var y: number = (y <<= x)')
+           .to.equal('num y = (y = (y as int) << (x as int));');
+
+       expectTranslate('var y: number = (y >>= x)')
+           .to.equal('num y = (y = (y as int) >> (x as int));');
+     });
+  it('should cast numeric values to int when bitwise data is evaluated as an input param into a function',
```
mprobst: I don't think this test adds value - it's just the regular operation of ts2dart, unrelated to your change. Please drop it.

test/expression_test.ts
```diff
@@ -92,4 +92,59 @@ describe('expressions', () => {
     expectTranslate('foo.bar;').to.equal('foo.bar;');
     expectTranslate('foo[bar];').to.equal('foo[bar];');
   });
+  it('should cast numeric values to int when bitwise operators are used for a LHS exp RHS numeric expression',
```
mprobst: Nit: LHS exp RHS ==> `binary expression`

mprobst: isn't this test largely redundant with the test table above? I'd drop it, doesn't add value.

mprobst:
A couple of nits left, but overall LGTMs.

Please clean up the tests, there's a lot of duplication in them that doesn't test anything that's not covered by other tests already.

test/expression_test.ts
```diff
@@ -92,4 +92,28 @@ describe('expressions', () => {
     expectTranslate('foo.bar;').to.equal('foo.bar;');
     expectTranslate('foo[bar];').to.equal('foo[bar];');
   });
+  it('should cast numeric values to int when bitwise operators are used for a binary expression',
```
mprobst: please remove this test, it's entirely redundant with the binary operator table test above.

test/expression_test.ts
```diff
@@ -92,4 +92,28 @@ describe('expressions', () => {
...
+     () => {
+       expectTranslates({
+         'var z: number = x | y': 'num z = (x as int) | (y as int);',
+         'var z: number = x & y': 'num z = (x as int) & (y as int);',
+         'var z: number = x ^ y': 'num z = (x as int) ^ (y as int);',
+         'var z: number = x << y': 'num z = (x as int) << (y as int);',
+         'var z: number = x >> y': 'num z = (x as int) >> (y as int);'
+       });
+     });
+  it('should cast numeric values to int when a shorthand bitwise assignment expression is used',
```
mprobst: please remove this test, it's entirely redundant with the binary operator table test above.

test/expression_test.ts
```diff
@@ -92,4 +92,28 @@ describe('expressions', () => {
...
+     () => {
+       expectTranslates({
+         'var y: number = (y |= x)': 'num y = (y = (y as int) | (x as int));',
+         'var y: number = (y &= x)': 'num y = (y = (y as int) & (x as int));',
+         'var y: number = (y ^= x)': 'num y = (y = (y as int) ^ (x as int));',
+         'var y: number = (y <<= x)': 'num y = (y = (y as int) << (x as int));',
+         'var y: number = (y >>= x)': 'num y = (y = (y as int) >> (x as int));',
+       });
+     });
+  it('should cast a number to int when evaluating bitwise not operators', () => {
```
mprobst: please remove this test, it's entirely redundant with the binary operator table test above.

mprobst:
Merged as ac5ac86, 97197d7.

PR #363, opened 2016-04-20
feat: ignore return types on function expressions.

### LINKED ISSUE #355: FunctionExpression assignment with explicit return type fails to transpile
Using version 0.7.27

The following TS code fails to transpile:

```ts
export function main() {
  var times2 = function(x: number): number {
    return x * 2;
  };
  console.log('5 * 2 = %s', times2(5));
}
```

Attempting to transpile this yields the following output:

```
fn.ts:1:1: Could not format because the source could not be parsed:

line 1, column 36: Expected to find ')'
 main ( ) { var times2 = num ( num x ) { return x * 2 ; } ; console . log ( "5 * 2 = %s" , times2 ( 5 ) ) ; }
                                   ^
line 1, column 32: Expected to find ';'
 main ( ) { var times2 = num ( num x ) { return x * 2 ; } ; console . log ( "5 * 2 = %s" , times2 ( 5 ) ) ; }
                               ^^^
line 1, column 36: Expected to find ';'
 main ( ) { var times2 = num ( num x ) { return x * 2 ; } ; console . log ( "5 * 2 = %s" , times2 ( 5 ) ) ; }
                                   ^
line 1, column 38: Expected an identifier
 main ( ) { var times2 = num ( num x ) { return x * 2 ; } ; console . log ( "5 * 2 = %s" , times2 ( 5 ) ) ; }
                                     ^
line 1, column 38: Expected to find ';'
 main ( ) { var times2 = num ( num x ) { return x * 2 ; } ; console . log ( "5 * 2 = %s" , times2 ( 5 ) ) ; }
                                     ^
line 1, column 38: Unexpected token ')'
 main ( ) { var times2 = num ( num x ) { return x * 2 ; } ; console . log ( "5 * 2 = %s" , times2 ( 5 ) ) ; }
```

Note that explicitly specifying `times2`s type still fails to compile.

This _does_ compile, however, if:

* You remove the explicit return type from `times2`.
* You turn `times2` into a declaration rather than an expression.

mprobst:
Same as for arrow functions, Dart does not have a syntax for return types.

Fixes #355.

### DIFF
--- a/lib/declaration.ts
+++ b/lib/declaration.ts
@@ -255,8 +255,10 @@ export default class DeclarationTranspiler extends base.TranspilerBase {
     this.fc.pushTypeParameterNames(fn);
     try {
       if (fn.type) {
-        if (fn.kind === ts.SyntaxKind.ArrowFunction) {
-          // Type is silently dropped for arrow functions, not supported in Dart.
+        if (fn.kind === ts.SyntaxKind.ArrowFunction ||
+            fn.kind === ts.SyntaxKind.FunctionExpression) {
+          // The return type is silently dropped for function expressions (including arrow
+          // functions), it is not supported in Dart.
           this.emit('/*');
           this.visit(fn.type);
           this.emit('*/');
--- a/test/function_test.ts
+++ b/test/function_test.ts
@@ -41,6 +41,12 @@ describe('functions', () => {
     expectTranslate('var a = (p = null) => isBlank(p)')
         .to.equal('var a = ([p = null]) => isBlank(p);');
   });
+  it('translates types on function expressions', () => {
+    expectTranslate('let a = function(p: string): string { return p; };')
+        .to.equal(`var a = /* String */ (String p) {
+  return p;
+};`);
+  });
   it('supports function parameters', () => {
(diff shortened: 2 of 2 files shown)


mprobst:
@traviskaufman, would you mind reviewing?

lib/declaration.ts
```diff
@@ -255,8 +255,10 @@ export default class DeclarationTranspiler extends base.TranspilerBase {
...
     try {
       if (fn.type) {
-        if (fn.kind === ts.SyntaxKind.ArrowFunction) {
-          // Type is silently dropped for arrow functions, not supported in Dart.
+        if (fn.kind === ts.SyntaxKind.ArrowFunction ||
+            fn.kind === ts.SyntaxKind.FunctionExpression) {
+          // The return type is silently dropped for function expressions (including arrow
+          // functions), it is not supported in Dart.
           this.emit('/*');
           this.visit(fn.type);
```
traviskaufman: will this work if the type is omitted? e.g. `let a = function(p) { return p; };`

mprobst: See the outer `if (fn.type)`, this never happens if there's no type at all. There's also a test for that.

traviskaufman: gotcha.

traviskaufman:
LGTM 👍

Issue #352, opened 2016-03-02
Support `console.log`


tbosch:
This should result in a `print` statement in Dart with interpolation.
E.g.
```
console.log('Name:', name)
```

Should be transpiled into
```
print(`Name: ${name}`);
```

jacob314:
dart:html could be easily updated to allow logging multiple arguments.
then the transformation would just need to be
console.log('Name:', name)
to 
window.console.log('Name:', name)

PR #364, opened 2016-04-22
feat: transpile console.log to print (#352)

ochafik:
Note: as suggested in the issue, we could transpile to
window.console.log if it had a varargs signature. However, that would
introduce a dependency to dart:html, which seems overkill.

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -506,6 +506,17 @@ export class FacadeConverter extends base.TranspilerBase {
       this.emit('is List');
       this.emit(')');
     },
+    'Console.log': (c: ts.CallExpression, context: ts.Expression) => {
+      this.emit('print(');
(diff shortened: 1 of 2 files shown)


lib/facade_converter.ts
```diff
@@ -421,6 +421,17 @@ export class FacadeConverter extends base.TranspilerBase {
   };
 
   private stdlibHandlers: ts.Map<CallHandler> = merge(this.es6Promises, {
+    'Console.log': (c: ts.CallExpression, context: ts.Expression) => {
```
mprobst: would you mind keeping these alphabetical?

mprobst:
LGTM.

lib/facade_converter.ts
ochafik: done, thanks!

Issue #194, opened 2015-06-14
RegExp.exec()


vicb:
`RegExp.exec()` get translated to `RegExp.allMatches()` I'm not sure if this is the right thing to do ?

In JS, when the regexp is global, you call `exec()` multiple times to get successive matches while Dart's `allMatches()` returns an ietrable of all macthes.

I don't think there is a 1 to 1 mapping here or may be I miss something ?

/cc @mprobst

mprobst:
Assigning vicb to come up with a good strategy on how to transpile this.

PR #365, opened 2016-04-23
Fix RegExp.exec (#194)

ochafik:
- Limit transpilation to direct call on a regexp literal (otherwise, side-effects of multiple calls on a global regexp can't be easily replicated with Dart's RegExp; a quick visitation of the tree could tell us if a regexp is non-global or if `exec` is only called once on it, but that would be overkill so we're just backing out of `x.exec(...)` if `x` is not a `/literal/`)
- Optimize case where `exec` result is only index-accessed

### DIFF
--- a/lib/facade_converter.ts
+++ b/lib/facade_converter.ts
@@ -517,14 +517,38 @@ export class FacadeConverter extends base.TranspilerBase {
       }
       this.emit(')');
     },
+    'RegExp.exec': (c: ts.CallExpression, context: ts.Expression) => {
+      if (context.kind !== ts.SyntaxKind.RegularExpressionLiteral) {
+        // Fail if the exec call isn't made directly on a regexp literal.
+        // Multiple exec calls on the same global regexp have side effects
+        // (each return the next match), which we can't reproduce with a simple
+        // Dart RegExp (users should switch to some facade / wrapper instead).
+        this.reportError(
+            c, 'exec is only supported on regexp literals, ' +
+                'to avoid side-effect of multiple calls on global regexps.');
+      }
+      if (c.parent.kind === ts.SyntaxKind.ElementAccessExpression) {
+        // The result of the exec call is used for immediate indexed access:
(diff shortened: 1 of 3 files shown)


lib/facade_converter.ts
```diff
@@ -511,9 +511,15 @@ export class FacadeConverter extends base.TranspilerBase {
...
     },
     'RegExp.exec': (c: ts.CallExpression, context: ts.Expression) => {
-      this.visit(context);
-      this.emitMethodCall('allMatches', c.arguments);
-      this.emitMethodCall('toList');
+      if (c.parent.kind === ts.SyntaxKind.ElementAccessExpression) {
+        this.visit(context);
+        this.emitMethodCall('firstMatch', c.arguments);
+      } else {
+        this.emit('((match) => new List.generate(match.groupCount + 1, match.group))(');
```
mprobst: This code is non-obvious, could you add some text explaining this?

test/facade_converter_test.ts
```diff
@@ -232,6 +232,15 @@ var y = x.length;`);
   var x = new RegExp(r'a');
   x.hasMatch("a");
 }`);
+    expectWithTypes('function f() { var x = /a(.)/g; x.exec("ab")[1]; }').to.equal(`f() {
+  var x = new RegExp(r'a(.)');
```
mprobst: Given that this was wrong before, could you add an end-to-end test for this?

lib/facade_converter.ts
ochafik: Done

ochafik:
Now forbidding exec call on non-literals (since we can't easily replicate the multiple-calls-on-same-global-regexp case).

test/facade_converter_test.ts
ochafik: Actually, now fixed the existing e2e test that initially came up red in travis :-)

mprobst:
LGTM.

PR #367, opened 2016-04-26
fix: Future.{value,error} are named constructors, not statics

ochafik:


mprobst:
LGTM.

mprobst:
@ochafik FYI we sent you an invite to become a member of ts2dart, so that you can merge this commit yourself. Welcome on board :-)

ochafik:
Thanks Martin!

PR #372, opened 2016-08-12
Add ability to install globally and execute as command line app

luisvt:


mprobst:
Thanks for the contribution!

I think the build failure might be a fluke (pub install something something), rerunning tests now to see if that's true.

mprobst:
I see, I think you might have to just bump the version of "test" in this file to `0.12.15`:
`https://github.com/angular/ts2dart/blob/master/test/e2e/pubspec.yaml`

Sorry for the yak shaving required here, this project has been somewhat less than active for a while, so bit rot creeps in.

luisvt:
Thanks, I will do it

mprobst:
Thanks, merged!

mprobst:
This has been released as `0.9.11`.

